PR #13, opened 2018-04-20
Client.list_dataset_tables(dataset) has been deprecated in favor of Client.list_tables(dataset)

amodig:
Hi,

There has been a method name change. Currently the master doesn't work with e.g. Superset. There was also an error in the docs that has been fixed:
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4712


mxmzdlv:
Thanks!

PR #16, opened 2018-08-10
Add support for STRUCT and ARRAY types.

tswast:
Also called RECORD and REPEATED fields.

Closes https://github.com/mxmzdlv/pybigquery/issues/4

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,8 +70,3 @@ Set up an environment and run tests::
     source .env/bin/activate
     pip install -r dev_requirements.txt
     pytest
-
-TODO
-====
-
-- Support for Record column type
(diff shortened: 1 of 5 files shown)


tswast:
Note: I created this from the GitHub web UI. Ideally we'd add some tests for these queries to make sure this actually fixes the issue.

sumedhsakdeo:
I check out this PR, build a python whl, and tested it locally against a BQ table with nested fields. And it works!
Thank you so much!

tswast:
I'm adding some tests now. I need to fix one thing for arrays.

tswast:
@sumedhsakdeo I'm glad the fix worked for you. Thanks for trying it out!

I believe this PR is ready for review now. I've added tests and updated the logic to also work for REPEATED / ARRAY types.

mxmzdlv:
This is fantastic. Thank you @tswast!

PR #18, opened 2018-08-17
Add support for nested columns

### LINKED ISSUE #17: bigquery record type handling is incomplete
Two issues: @tswast  @mxmzdlv 
1. get_columns does not recurse into record type and return all the nested field names.
    -- I have a PR for this. Will post it for review. Also attached a working code here.
2. when you do a select \`abc.xyz\` from \`foobar-table\`
    -- bigquery does not like this syntax for the column name put in backtick, like here \`abc.xyz\`

Need recommendation on how to handle issue (2) as it blocks running tests and hence fix to issue (1).

Here is the diff for issue (1) which correctly returns all the column names
```
diff --git a/pybigquery/sqlalchemy_bigquery.py b/pybigquery/sqlalchemy_bigquery.py
index 071ddfd..1c31f9e 100644
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals

 from google.cloud.bigquery import dbapi
+from google.cloud.bigquery.schema import SchemaField
 from google.api_core.exceptions import NotFound
 from sqlalchemy.exc import NoSuchTableError
@@ -160,9 +161,24 @@ class BigQueryDialect(DefaultDialect):
         except NoSuchTableError:
             return False

+    def _get_columns_helper(self, columns, cur_columns):
+        results = []
+        for col in columns:
+            if col.field_type == 'RECORD':
+                cur_columns.append(col)
+                results += self._get_columns_helper(col.fields, cur_columns)
+                cur_columns.pop()
+            results += [SchemaField(
+                            name='.'.join(col.name for col in cur_columns + [col]),
+                            field_type=col.field_type,
+                            mode=col.mode,
+                            description=col.description,
+                            fields=col.fields)]
+        return results
+
     def get_columns(self, connection, table_name, schema=None, **kw):
         table = self._get_table(connection, table_name, schema)
-        columns = table.schema
+        columns = self._get_columns_helper(table.schema, [])
         result = []
         for col in columns:
             try:
```

mxmzdlv:
Fixes #17

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -129,8 +129,6 @@ class BigQueryCompiler(SQLCompiler):
         args[0].use_labels = True
         return super(BigQueryCompiler, self).visit_select(*args, **kwargs)
 
-
-
     def visit_column(self, column, add_to_result_map=None,
                      include_table=True, **kwargs):
 
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 from google.cloud.bigquery import dbapi
+from google.cloud.bigquery.schema import SchemaField
 from google.cloud import bigquery
 from google.api_core.exceptions import NotFound
 from sqlalchemy.exc import NoSuchTableError
@@ -12,6 +13,7 @@ from sqlalchemy.sql.compiler import SQLCompiler, IdentifierPreparer
 from sqlalchemy.engine.default import DefaultDialect, DefaultExecutionContext
 from sqlalchemy.engine.base import Engine
 from sqlalchemy.sql.schema import Column
+from sqlalchemy.sql import elements
 
 
 class UniversalSet(object):
@@ -36,6 +38,36 @@ class BigQueryIdentifierPreparer(IdentifierPreparer):
             dialect,
             initial_quote="`",
         )
+
+    def quote_column(self, value):
+        """
+        Quote a column.
+        Fields are quoted separately from the record name.
+        """
+
+        parts = value.split('.')
+        return '.'.join(self.quote_identifier(x) for x in parts)
+
+    def quote(self, ident, force=None, column=False):
+        """
+        Conditionally quote an identifier.
+        """
+
+        force = getattr(ident, "quote", None)
+
+        if force is None:
+            if ident in self._strings:
+                return self._strings[ident]
+            else:
+                if self._requires_quotes(ident):
+                    self._strings[ident] = self.quote_column(ident) if column else self.quote_identifier(ident)
+                else:
+                    self._strings[ident] = ident
+                return self._strings[ident]
+        elif force:
+            return self.quote_column(ident) if column else self.quote_identifier(ident)
+        else:
+            return ident
 
     def format_label(self, *args, **kwargs):
         # Replace dots with underscores in labels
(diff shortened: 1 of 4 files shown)


pybigquery/sqlalchemy_bigquery.py
```diff
@@ -160,9 +233,28 @@ def has_table(self, connection, table_name, schema=None):
...
             return False
 
+    def _get_columns_helper(self, columns, cur_columns):
+        """
+        Recurse into record type and return all the nested field names.
+        As contributed by @sumedhsakdeo on issue #17
+        """
+        results = []
+        for col in columns:
+            if col.field_type == 'RECORD':
```
tswast: Please add `and col.mode != 'REPEATED'` to this line. It's not possible to select nested columns in an array of structures without extra calls to `UNNEST`

sumedhsakdeo: @tswast sure. good point.

Thinking more, should we change this code to ignore the name of the record field. And return only leaf field names in the record field.
```
+           if col.field_type == 'RECORD' and col.mode != 'REPEATED':
+                cur_columns.append(col)
+                results += self._get_columns_helper(col.fields, cur_columns)
+                cur_columns.pop()
+            else: <<<<<<<<
+                 results += [SchemaField(
+                            name='.'.join(col.name for col in cur_columns + [col]),
+                            field_type=col.field_type,
+                            mode=col.mode,
+                            description=col.description,
+                            fields=col.fields)]
```

sumedhsakdeo:
It is also important to use a label that is unique for each record type field column
See my PR on how it handle these issues with column names and their labels. https://github.com/apache/incubator-superset/pull/5655

pybigquery/sqlalchemy_bigquery.py
tswast: > return only leaf field names in the record field.

Makes sense to me. It'd probably be weird to have both the record fields and the subfields.

sumedhsakdeo: @mxmzdlv could you please incorporate this diff in the _get_columns_helper(). Thanks.

mxmzdlv:
Added the requested changes. I've tested this with Superset and it seems to be working fine. Can anyone please test as well?

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -12,6 +13,11 @@
 from sqlalchemy.engine.default import DefaultDialect, DefaultExecutionContext
 from sqlalchemy.engine.base import Engine
 from sqlalchemy.sql.schema import Column
+from sqlalchemy.sql import elements
+import re
+
+
+FIELD_LEGAL_CHARACTERS = re.compile('[^\w+r]+')
```
sumedhsakdeo: should this variable be named FIELD_ILLEGAL_CHARACTERS as per the regex definition?
Also, why is this not sufficient `[^\w]+`?
Other criteria for legal field name is it must beginning with alphabet or underscore.
https://github.com/apache/incubator-superset/pull/5641/files

sumedhsakdeo:
@mxmzdlv thanks for this PR, appreciate it. I will test this PR with superset and BQ tables with nested columns on Sunday or early Monday (P.T.), and get back.

sumedhsakdeo:
It works, I tested with Apache Superset and BQ. 🎉 thank you. @mxmzdlv

pybigquery/sqlalchemy_bigquery.py
mxmzdlv: You are right, thanks.

For the first character I am doing a check on line 79 and append underscore if it is invalid:

```python
if not name[0].isalpha() and name[0] != '_':
    name = "_" + name
```

PR #19, opened 2018-08-21
Fix pypi long_description

mistercrunch:
screen shot 2018-08-20 at 10 56 52 pm


mxmzdlv:
Thanks @mistercrunch!

PR #20, opened 2018-08-22
Google BigQuery API Client

sumedhsakdeo:
We would like the ability to support bigquery's dryrun queries in superset. For this, we would have to use BigQuery API. Introducing an API client in the pybigquery library so that clients like superset can use the APIClient in this library.

@mxmzdlv @tswast wdyt about this addition?

### DIFF
--- a/pybigquery/api.py
+++ b/pybigquery/api.py
@@ -1,0 +1,22 @@
+"""Integration with BigQuery API."""
+
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+from google.cloud.bigquery import Client, QueryJobConfig
+
+
+class ApiClient(object):
+    def __init__(self, credentials_path=None):
+        self.credentials_path = credentials_path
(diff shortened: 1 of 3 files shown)


pybigquery/api.py
```diff
@@ -0,0 +1,22 @@
+"""Integration with BigQuery API."""
+
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+from google.cloud.bigquery import Client, QueryJobConfig
+
+
+class ApiClient(object):
+    def __init__(self, credentials_path=None):
```
tswast: I'd prefer if this took a `google-auth` library credentials object. That way it can support other kinds of credentials such as Compute Engine credentials or user credentials.

tswast: But this is okay to me if Superset expects a path here.

sumedhsakdeo: Yes we use path in superset

PR #22, opened 2018-09-05
py3 compatible README.rst parsing

mistercrunch:
Hit error `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in
position 780: ordinal not in range(128)` on py3.

Tested this solution against python 3.6 and 2.7


mxmzdlv:
Thanks!

PR #23, opened 2018-09-10
Allow create_engine to receive 'credentials_info' python dict

mistercrunch:
This provides an alternative to the JSON `credentials_path` approach.
This allows passing a `credentials_info` dict when calling
`create_engine`.

My specific use-case is around using `DB_CONNECTION_MUTATOR`
configuration hook in Apache Superset to impersonate the right BQ user at query
time. At that moment we'll be altering the `credentials_info` with
`serviceAccountUser` pointing to the proper user instead of using the
service account.


mxmzdlv:
👍

PR #25, opened 2018-09-19
adding support for datasets and query arguments in connection string

### LINKED ISSUE #24: Add ability to specify dataset in connection string
We have a use case where we want to treat separate datasets as similar to separate databases or schemas in a traditional database. This would allow us to have separate bindings to different datasets.

An example of the kind of behavior we'd like:

```py
engine = create_engine('bigquery://some-project-name/some_dataset')
```

This is a pretty important use case for us, since it would make a multi-tenant approach much easier.

blainehansen:
closes #24

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,10 +79,25 @@ If you want to have the ``Client`` use a default dataset, specify it as the "dat
 
     engine = create_engine('bigquery://project/dataset')
 
+When using a default dataset, don't include the dataset name in the table name, e.g.:
+
+.. code-block:: python
+
+    table = Table('table_name')
+
+Note that specyfing a default dataset doesn't restrict execution of queries to that particular dataset when using raw queries, e.g.:
+
+.. code-block:: python
+
+    # Set default dataset to dataset_a
+    engine = create_engine('bigquery://project/dataset_a')
+
+    # This will still execute and return rows from dataset_b
+    engine.execute('SELECT * FROM dataset_b.table').fetchall()
+
 
 Connection String Parameters
 ____________________________
-
 
 There are many situations where you can't call ``create_engine`` directly, such as when using tools like `Flask SQLAlchemy <http://flask-sqlalchemy.pocoo.org/2.3/>`_. For situations like these, or for situations where you want the ``Client`` to have a `default_query_job_config <https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client>`_, you can pass many arguments in the query of the connection string.
 
--- a/README.rst
+++ b/README.rst
@@ -84,9 +84,9 @@ Connection String Parameters
 ____________________________
 
 
-There are many situations where you can't call ``create_engine`` directly, such as when using tools like `Flask SQLAlchemy <http://flask-sqlalchemy.pocoo.org/2.3/>`_. For situations like these, or for situations where you want the ``Client`` to have a ```default_query_job_config`` <https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client>`_, you can pass many arguments in the query of the connection string.
+There are many situations where you can't call ``create_engine`` directly, such as when using tools like `Flask SQLAlchemy <http://flask-sqlalchemy.pocoo.org/2.3/>`_. For situations like these, or for situations where you want the ``Client`` to have a `default_query_job_config <https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client>`_, you can pass many arguments in the query of the connection string.
 
-The ``credentials_path``, ``location``, and ``arraysize`` parameters are used by this library, and the rest are used to create a ```QueryJobConfig`` <https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.job.QueryJobConfig.html#google.cloud.bigquery.job.QueryJobConfig>`_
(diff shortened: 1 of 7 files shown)


blainehansen:
This has only been tested with a basic case, a simply modified version of the example on the readme:

```py
from sqlalchemy import *
from sqlalchemy.engine import create_engine
from sqlalchemy.schema import *
engine = create_engine('bigquery://some_project/some_dataset')

table = Table('some_table', MetaData(bind=engine), autoload=True)
print(select([func.count('*')], from_obj=table).scalar())
```

This will successfully count the rows in `some_dataset.some_table`

blainehansen:
I'm not sure what other modifications would need to be made, or if this could be achieved in a better way.

mxmzdlv:
Thanks @blainehansen!

I've tested the code and there are some issues with SQL generation, specifically when generating queries with GROUP, HAVING or ORDER labels.

With your approach table names don't include dataset ids, and in that case GROUP BY / HAVING / ORDER should not include them either, but they are still included in the generated query:

E.g., this is what SQLAlchemy generates:
```
SELECT `sample`.`string` AS `sample_string`, count(`sample`.`integer`) AS `count_1` 
FROM `test_pybigquery`.`sample` 
GROUP BY `test_pybigquery`.`sample`.`string`
```

And it needs to be:
```
SELECT `sample`.`string` AS `sample_string`, count(`sample`.`integer`) AS `count_1` 
FROM `test_pybigquery`.`sample`
GROUP BY `sample`.`string`
```

Also, `get_table_names` method still includes dataset id in it and I think that would cause issues with auto-generated table definitions by Superset.

I reckon we should try to simplify this and still try to require dataset id in the table name. This will  also make everything more consistent.

Some of the relevant tests that you should look at are `test_tables_list`, `test_group_by`.

mxmzdlv:
Another issue is that the underlying BigQuery client doesn't allow restricting query execution per specified dataset, so raw queries like that will still execute on any dataset:

```
engine.execute('SELECT * FROM another_dataset.table').fetchall()
```

blainehansen:
**To your first comment:**

We're completely fine if the `dataset_id` is appended everywhere (as in your last example):

```sql
SELECT `test_pybigquery.sample`.`string` AS `test_pybigquery_sample_string`, count(`test_pybigquery.sample`.`integer`) AS `count_1` 
FROM `test_pybigquery.sample`
GROUP BY `test_pybigquery.sample`.`string`
```

How do I change the code to make that possible? I'm not sure where I can take the `url.database` string (which is where it's made available from the connection string), and input it to make it included everywhere, without simply naming the table `dataset_id.table_name` (that would obviously ruin the multi-tenancy). The ideal solution for us is to change your codebase as little as possible to make this feature happen.
(sorry if I seem a little clueless, I've had a very difficult time finding useful docs for sqlalchemy custom dialects, even yesterday finding out about the `visit_table` function was a long process of searching across random sites like kite.com and then guessing and checking. I'd love your help or pointers in getting this figured out, whether it's code suggestions or pointers to good resources)


**To your second comment:**

Yeah, raw queries acting on random datasets (or just generally not having any awareness of the dataset restriction) is definitely a problem for us. But without changing the bigquery client to restrict to a specific dataset, I'm not sure how that could be solved. We really want to avoid having to create separate projects for different tenants, that promises to have a lot of maintenance overhead.


**Thanks!**

I'll start playing with the code with your tests to see what I can make happen.

mxmzdlv:
I think you would need to modify `BigQueryCompiler.visit_label` or `BigQueryIdentifierPreparer.format_label`, and `include_table` in `BigQueryCompiler` might be relevant, but I am not sure how to get dataset_id into these methods. Perhaps you could have another BigQueryCompiler with a separate logic for when the dataset is specified on connection — you could check it in `BigQueryDialect.create_connect_args` and overwrite `statement_compiler`.

I haven't had much luck with the documentation either, so I generally just look into source code — 
BigQueryDialect inherits from [DefaultDialect](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/engine/default.py#L36) and SQL generation mostly happens in [SQLCompiler and IdentifierPreparer](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/compiler.py).

blainehansen:
The pull request at https://github.com/GoogleCloudPlatform/google-cloud-python/pull/6088 is actually looking like the maintainers are interested in going in that direction, so I think this will get much easier to achieve once you can pass a default `QueryJobConfig` to the `Client` constructor. I'm going to change the direction of this pull request to accept more default config parameters in the [sqlalchemy `URL`](https://docs.sqlalchemy.org/en/latest/core/engines.html#sqlalchemy.engine.url.URL) to create that default `QueryJobConfig`.

mxmzdlv:
Awesome! I've added some additional tests and changed `get_schema_names` and `get_table_names` behavior when dataset is set. Seems to be passing all the tests. I've also tested it with Superset and haven't seen any issues.

Since this still won't prevent executing queries against other datasets when using engine directly — I've added a note regarding that in readme.

I think this is good to merge as soon as the new version of google-cloud-python lands.

blainehansen:
[That landed about 16 hours ago](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/6088), so it's possible that the tests you've run have already validated everything! Let me know if there's anything else I can do to help.

mxmzdlv:
Yeah, I've manually added changes from that pull request. 

Ideally we should wait for the release, but if you need to use this ASAP you can try pointing to the github repo in setup.py and then do `pip install https://github.com/marketdial/pybigquery/archive/master.zip` in your project.

blainehansen:
Alright, the bigquery code has been released!
https://github.com/googleapis/google-cloud-python/releases

mxmzdlv:
Perfect! Pushed the new version 0.4.6.

blainehansen:
:confetti_ball: :confetti_ball: :confetti_ball: 

Thanks!

PR #27, opened 2018-10-02
add support for numeric types in pybigquery

sumedhsakdeo:
CC @tswast @mistercrunch @mxmzdlv

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -99,7 +99,7 @@ _type_map = {
     'BYTES': types.BINARY,
     'TIME': types.TIME,
     'RECORD': types.JSON,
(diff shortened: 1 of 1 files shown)


pybigquery/sqlalchemy_bigquery.py
```diff
@@ -99,6 +99,7 @@ def format_label(self, label, name=None):
     'BYTES': types.BINARY,
     'TIME': types.TIME,
     'RECORD': types.JSON,
+    'NUMERIC': types.Numeric,
```
tswast: Looks like Numeric is a base class. We might prefer `sqlalchemy.types.DECIMAL` instead.

sumedhsakdeo: Yes, good point. Done.

PR #29, opened 2018-10-24
fix get column names for repeated fields of type record

sumedhsakdeo:
Currently, getColumns for repeated fields of type record only returns the top level column name.

Example, 
Instead of the returning something like,
```
array_field_name type:Record, mode:Repeated
array_field_name.foo : string
array_field_name.bar : string
```

getColumns() is returning 
```
array_field_name type:Record, mode:Repeated`
```

>> This PR is trying to address the above bug.

For type:record that is a non-repeated field, we will return something like this,
```
array_field_name type:Record, mode:NULLABLE
array_field_name.foo : string
array_field_name.bar : string
```

CC: @tswast @mistercrunch @mxmzdlv 
let me know if this is OK?

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -311,16 +311,15 @@ class BigQueryDialect(DefaultDialect):
         """
         results = []
         for col in columns:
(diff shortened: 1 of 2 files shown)


mxmzdlv:
LGTM

PR #32, opened 2018-12-18
Add type compiler.

jmcarp:
Map generic datatypes to bigquery datatypes.

With this patch, I was able to create tables with common types via pybigquery.

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -12,7 +12,7 @@ from google.oauth2 import service_account
 from google.api_core.exceptions import NotFound
 from sqlalchemy.exc import NoSuchTableError
 from sqlalchemy import types, util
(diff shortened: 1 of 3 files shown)


tswast:
Neat!

I know we have a dictionary from BigQuery types to SQLAlchemy types. Would it make sense to invert that dictionary and use it for these methods? I hate to have the same mapping defined in two places, but I guess this is the only place so far where we need to go from SQLAlchemy types to BigQuery types.

jmcarp:
Amended to add a few more common types.

jmcarp:
I was testing with sqlalchemy core, which apparently does support tables without primary keys. For sqlalchemy declarative, tables without primary keys [aren't supported](https://docs.sqlalchemy.org/en/latest/faq/ormconfiguration.html#how-do-i-map-a-table-that-has-no-primary-key), but I added a ddl compiler that skips foreign and primary keys on table creation.

mxmzdlv:
Thanks!

PR #33, opened 2019-01-06
Support table and column metadata.

jmcarp:
To add metadata to a table:

```python
table = Table('mytable', ..., bigquery_description='my table description')
```

To add metadata to a column:

```python
Column('mycolumn', doc='my column description')
```


jmcarp:
Hopefully nobody is accepting untrusted user input for DML statements. Since we're quoting options, I think this has the same risk of injection as table and column names.

PR #44, opened 2019-12-03
Solve issue 43, Fix default dataset

alonme:
Author:    alonme <alon.menczer@gmail.com>

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+from google import auth
 from google.cloud import bigquery
 from google.cloud.bigquery import dbapi, QueryJobConfig
 from google.cloud.bigquery.schema import SchemaField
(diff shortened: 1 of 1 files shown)


pybigquery/sqlalchemy_bigquery.py
```diff
@@ -267,47 +268,61 @@ def __init__(
...
+        if dataset_id:
+            # If project_id is missing, use defualt project_id
+            if not project_id:
+                _, project_id = auth.default()
 
-        self.arraysize = self.arraysize or arraysize
-        self.location = location or self.location
-        self.credentials_path = credentials_path or self.credentials_path
-        self.dataset_id = dataset_id
+            job_config.default_dataset = f'{project_id}.{dataset_id}'
```
tswast: What versions of Python does `pybigquery` want to support? If we need to support Python 3.5, I don't think we can use fstrings.

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -267,47 +268,61 @@ def __init__(
...
-            credentials = credentials.with_scopes(scopes)
-            client = bigquery.Client(
-                project=self.credentials_info.get('project_id'),
+        credentials = credentials.with_scopes(scopes)
+
+        self._add_default_dataset_to_job_config(default_query_job_config,
+                                        credentials.project_id, self.dataset_id)
+
+        return bigquery.Client(
+                project=credentials.project_id,
```
tswast: Note: there are some kinds of credentials that don't have a `project_id` property. e.g. if someone uses `gcloud application-default login`.

Edit: Looks like it's not relevant, as this is only called with service account credentials.

pybigquery/sqlalchemy_bigquery.py
alonme: Resolved

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -267,47 +268,61 @@ def __init__(
     def dbapi(cls):
         return dbapi
 
-    def create_connect_args(self, url):
-        location, dataset_id, arraysize, credentials_path, default_query_job_config = parse_url(url)
+    @staticmethod
+    def _add_default_dataset_to_job_config(job_config, project_id, dataset_id):
+        # If dataset_id is set, then we know the job_config isn't None
+        if dataset_id:
+            # If project_id is missing, use defualt project_id
```
tswast:
```suggestion
            # If project_id is missing, use default project_id
```

alonme:
@mxmzdlv Can you kindly merge and deploy a new version for this bug fix?

mxmzdlv:
Thanks for fixing this!

PR #45, opened 2020-01-04
Fix NUMERIC DDL representation.

jmcarp:
BigQuery NUMERIC types don't have configurable scale or precision, so
override the default visitor to drop unused arguments.


jmcarp:
@mxmzdlv would appreciate a review when you get time.

mxmzdlv:
Looks good, thanks!

PR #49, opened 2020-02-19
Set BigQueryDialect.supports_multiline_insert to True

jmaliksi:
The default sqlalchemy `visit_insert` [compiler method](https://github.com/sqlalchemy/sqlalchemy/blob/5675345fc11966d449130ff4e4327a5a4bece0c2/lib/sqlalchemy/sql/compiler.py#L2828) compiles a correct BigQuery INSERT statement mostly out-of-the-box, it just needs to have this flag enabled.


mxmzdlv:
Thanks!

Issue #48, opened 2020-02-11
project_id is ignored when using service account with different project id


mindartur:
Whatever options is used, can't override project ID from service account

tswast:
Looks like we need to update https://github.com/mxmzdlv/pybigquery/blob/1a76103a445de6ad9bb329346ecf075d1255a118/pybigquery/sqlalchemy_bigquery.py#L303 to pass the `url.host` to `self._create_client_from_credentials`.

Pull requests welcome.

alonme:
@tswast we can resolve this issue, as it is now merged in available in the release

tswast:
I add some unit tests for this in https://github.com/googleapis/python-bigquery-sqlalchemy/pull/100

PR #52, opened 2020-04-15
Use the provided project_id when using a service account

alonme:
solve #48


alonme:
@tswast do you have any ideas regarding how we can write tests that for the service account connection options? 
currently it is not tested at all.

alonme:
I'll add tests for that this weekend hopefully.
the connected issue can probably be closed.

alonme:
resolved #48

PR #61, opened 2020-06-26
ISSUE-41: Support storage api

### LINKED ISSUE #41: Support of Storage API
Any plan on supporting the bigquery storage api ?

---

**Edit** Sub-steps:

- [x] Update BQ DB-API to support BigQuery Storage API https://github.com/googleapis/python-bigquery/issues/16
- [ ] Update `create_connect_args` to (optionally) return a BQ Storage API client alongside the BQ client.
- [ ] Add a `use_bqstorage_api` option to the connection URL parsing logic.

alonme:
closes #41

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,7 +1,7 @@
 sqlalchemy>=1.1.9
-google-cloud-bigquery[bqstorage]>=1.25.0
-google-cloud-bigquery-storage[fastavro]>=1.0.0
+google-cloud-bigquery>=1.12.0
 future==0.16.0
+packaging>=20.0
 
 pytest==3.2.2
 pytz==2017.2
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,5 +1,6 @@
 sqlalchemy>=1.1.9
-google-cloud-bigquery>=1.6.0
+google-cloud-bigquery[bqstorage]>=1.25.0
+google-cloud-bigquery-storage[fastavro]>=1.0.0
 future==0.16.0
 
 pytest==3.2.2
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 from google import auth
-from google.cloud import bigquery
+from google.cloud import bigquery, bigquery_storage_v1beta1
 from google.cloud.bigquery import dbapi, QueryJobConfig
 from google.cloud.bigquery.schema import SchemaField
 from google.cloud.bigquery.table import EncryptionConfiguration
@@ -320,7 +320,7 @@ class BigQueryDialect(DefaultDialect):
         return credentials
 
     def create_connect_args(self, url):
-        project_id, location, dataset_id, arraysize, credentials_path, default_query_job_config = parse_url(url)
+        project_id, location, dataset_id, arraysize, credentials_path, use_bqstorage_api, default_query_job_config = parse_url(url)
 
         self.arraysize = self.arraysize or arraysize
         self.location = location or self.location
@@ -341,8 +341,12 @@ class BigQueryDialect(DefaultDialect):
             location=self.location,
             default_query_job_config=default_query_job_config
         )
+        
+        storage_client = None 
+        if use_bqstorage_api:
+            storage_client = bigquery_storage_v1beta1.BigQueryStorageClient(credentials=credentials)
 
-        return ([client], {})
+        return ([client, storage_client], {})
 
     def _json_deserializer(self, row):
         """JSON deserializer for RECORD types.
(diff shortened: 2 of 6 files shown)


dev_requirements.txt
```diff
@@ -1,5 +1,6 @@
 sqlalchemy>=1.1.9
-google-cloud-bigquery>=1.6.0
+google-cloud-bigquery[bqstorage]>=1.25.0
+google-cloud-bigquery-storage[fastavro]>=1.0.0
```
alonme, line 3: I am not sure what should be done concerning this.
`fastavro` was needed for the most basic functionality for me, and i think `pyarrow` may also be needed for other basic scenarios.
Should we include both as extras?

setup.py
```diff
@@ -25,7 +25,8 @@ def readme():
     ],
     install_requires=[
         'sqlalchemy>=1.1.9',
-        'google-cloud-bigquery>=1.6.0',
+        'google-cloud-bigquery[bqstorage]>=1.25.0',
+        'google-cloud-bigquery-storage[fastavro]>=1.0.0',
```
alonme, line 29: See dev_requirements.txt comment

alonme:
@tswast its a bit of a funny situation, we can't work with a storage api with `google-cloud-bigquery` < 1.25.0, and we can't use the `V1` with less then 1.26.
However, if the user does have 1.26, and the `bqstorage` extras, than the `use_bqstorage_api` flag is irrelevant as it is used by default in 1.26.

i will try and make it work without changing the current dependencies, but it feels to me like we should try and move to 1.26, if there are not many breaking changes between 1.12 and 1.26 (current is 1.6, but it seems we already break with anything below 1.12)

dev_requirements.txt
```diff
@@ -1,6 +1,7 @@
 sqlalchemy>=1.1.9
-google-cloud-bigquery>=1.6.0
+google-cloud-bigquery>=1.12.0
```
alonme, line 2: the module doesn't work with anything below that ( also in master)

tswast:
> its a bit of a funny situation, we can't work with a storage api with google-cloud-bigquery < 1.25.0, and we can't use the V1 with less then 1.26. However, if the user does have 1.26, and the bqstorage extras, than the use_bqstorage_api flag is irrelevant as it is used by default in 1.26.

@alonme I didn't realize that the DB-API is using the BigQuery Storage API by default now. That's actually really good news. I wonder if that means we can drop this PR altogether? Or do you think it'd be useful to be able to explicitly enable/disable this API usage?

alonme:
@tswast Its used by default for version 1.26 (which we don't require) and only if you install the extra dependencies.

The storage api has different pricing than the regular queries, so it kind of makes sense to me, however they are using it as default in the DB-API...

We will probably want to add an option to install the extras through this package,
Another question is if we should reflect the usage of the storage api to the user in any way? reflect that he misses dependencies needed to use it?

This PR also contains some refactoring and fixing a requirements issue - so if we decide to not use it - i'll open another one for these changes

tswast:
> The storage api has different pricing than the regular queries, so it kind of makes sense to me, however they are using it as default in the DB-API...

It's actually free to download results from query results tables and can be used with BigQuery Sandbox, so it shouldn't make a difference regarding pricing.

It is an extra API to enable and extra packages to install, so there will certainly be some users who aren't able to use it. 🤔

alonme:
So what do we want to do with this?

alonme:
So what do we want to do with this?

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -294,53 +302,67 @@ def _add_default_dataset_to_job_config(job_config, project_id, dataset_id):
...
+            default_query_job_config=default_query_job_config
+        )
+        
+        clients = [client]
+
+        if use_bqstorage_api:
+            if parse_version(bigquery.__version__) >= Version("1.26.0"):
+                try:
+                    storage_client = bigquery_storage_v1.BigQueryReadClient(credentials=credentials)
+                    clients.append(storage_client)
```
tswast, line 359: Since `bqstorage_client` is an optional argument to [`connect()`](https://googleapis.dev/python/bigquery/latest/dbapi.html#google.cloud.bigquery.dbapi.connect), we should probably be using the dictionary for this rather than appending to the list of positional arguments. (In fact, maybe we should do the same for the BigQuery client object, too)

tswast:
Closing, since there doesn't seem to be demand for turning _off_ the BigQuery Storage API support and that API is now used by default.

Thank you for your investigation and work on this @alonme!

gallodustin:
> Looking at the DB-API docs, it looks like it's not actually possible to explicitly disable the BQ Storage API with the DB-API. That makes setting an option to `False` a bit misleading. If we do want to proceed with this PR, we should file an issue on https://github.com/googleapis/python-bigquery to have a way to disable using the BQ Storage API.
> 
> Probably worth waiting to see if folks have trouble with the current default behavior before filing that, though.

Opened https://github.com/googleapis/python-bigquery/issues/1920.

PR #62, opened 2020-07-29
ISSUE-60: Implement get_view_names()

dimkonko:
Issue: https://github.com/mxmzdlv/pybigquery/issues/60

**Major change:** in this change `get_table_names` will return **only** table names (without view names).

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -3,6 +3,8 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import operator
+
 from google import auth
 from google.cloud import bigquery
 from google.cloud.bigquery import dbapi, QueryJobConfig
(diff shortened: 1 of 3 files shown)


tswast:
Makes sense, but does Superset call `get_view_names`? I'm worried that dashboarding tools might not see all the tables they're intended to see if `get_table_names` only returns non-view tables.

romainr:
Interesting: https://docs.sqlalchemy.org/en/13/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_table_names

`The names are expected to be real tables only, not views. Views are instead returned using the Inspector.get_view_names() method.`

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -349,6 +356,26 @@ def _json_deserializer(self, row):
...
+        get_table_name = self._build_formatted_table_id \
+            if self.dataset_id is None else \
+            operator.attrgetter("table_id")
+
+        client = connection.connection._client
+        datasets = client.list_datasets()
+
+        result = []
+        for dataset in datasets:
+            if current_schema is not None and current_schema != dataset.dataset_id:
```
tswast, line 370: Is there any reason to call `list_datasets` at all if `current_schema` is set?

tswast (approved):
Lovely! Thank you for adding tests as well.

romainr:
Thanks!

And for the record this is a backward incompatible change, so good to be explicit in the release notes

tswast:
@romainr Good point. I've updated the CHANGELOG and release notes.

PR #63, opened 2020-08-12
Allow use of schema argument for project and dataset

vinceatbluelabs:
This is a follow-on PR to https://github.com/mxmzdlv/pybigquery/pull/53 to implement the desired expanded project and dataset parsing behavior discussed during its review.

Tests have been updated and pass.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,11 +54,14 @@ To specify location of your datasets pass ``location`` to ``create_engine()``:
 Table names
 ___________
 
-To query tables from non-default projects, use the following format for the table name: ``project.dataset.table``, e.g.:
+To query tables from non-default projects or datasets, use the following format for the SQLAlchemy schema name: ``[project.]dataset``, e.g.:
 
 .. code-block:: python
 
-    sample_table = Table('bigquery-public-data.samples.natality')
+    # If neither dataset nor project are the default
+    sample_table_1 = Table('natality', schema='bigquery-public-data.samples')
+    # If just dataset is not the default
+    sample_table_2 = Table('natality', schema='bigquery-public-data')
 
 Batch size
 __________
@@ -85,7 +88,7 @@ When using a default dataset, don't include the dataset name in the table name, 
 
     table = Table('table_name')
 
-Note that specyfing a default dataset doesn't restrict execution of queries to that particular dataset when using raw queries, e.g.:
+Note that specifying a default dataset doesn't restrict execution of queries to that particular dataset when using raw queries, e.g.:
 
 .. code-block:: python
 
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -366,19 +366,24 @@ class BigQueryDialect(DefaultDialect):
 
         return (project, dataset, table_name)
 
+    def _table_reference(self, provided_schema_name, provided_table_name,
+                         client_project):
+        project_id, dataset_id, table_id = self._split_table_name(provided_table_name)
+        project_id = project_id or client_project
+        dataset_id = dataset_id or provided_schema_name or self.dataset_id
+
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -54,11 +54,14 @@ To specify location of your datasets pass ``location`` to ``create_engine()``:
...
-To query tables from non-default projects, use the following format for the table name: ``project.dataset.table``, e.g.:
+To query tables from non-default projects or datasets, use the following format for the SQLAlchemy schema name: ``[project.]dataset``, e.g.:
 
 .. code-block:: python
 
-    sample_table = Table('bigquery-public-data.samples.natality')
+    # If neither dataset nor project are the default
+    sample_table_1 = Table('natality', schema='bigquery-public-data.samples')
+    # If just dataset is not the default
+    sample_table_2 = Table('natality', schema='bigquery-public-data')
```
vinceatbluelabs, line 64: As discussed in https://github.com/mxmzdlv/pybigquery/pull/53, I believe it's more in the spirit of SQLAlchemy and common SQL database notions of 'schema name' to specify the location, rather than encouraging people to specify it in the name (in BigQuery-speak, the project and dataset).

As per previous discussion, the code will continue to accept dataset and project specified in the table name to avoid breaking existing users, and there are regression tests added for that usage.

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -363,22 +363,51 @@ def _split_table_name(full_table_name):
             dataset, table_name = table_name_split
         elif len(table_name_split) == 3:
             project, dataset, table_name = table_name_split
+        else:
+            raise ValueError("Did not understand table_name: {}".format(full_table_name))
```
vinceatbluelabs, line 367: I used .format() instead of f-strings since I assume you want to support older Python versions.  The underlying BigQuery libraries seemed to use ValueError for SQL parsing issues, so I this was a good exception type.

pybigquery/sqlalchemy_bigquery.py
```diff
@@ -363,22 +363,51 @@ def _split_table_name(full_table_name):
...
+        if (project_id_from_schema and project_id_from_table and
+           project_id_from_schema != project_id_from_table):
+            raise ValueError("project_id specified in schema and table_name disagree: got {} in schema, and {} in table_name".format(project_id_from_schema, project_id_from_table))
+        project_id = project_id_from_schema or project_id_from_table or client_project
+        dataset_id = dataset_id_from_schema or dataset_id_from_table or self.dataset_id
+
+        table_ref = TableReference.from_string("{}.{}.{}".format(
+            project_id, dataset_id, table_id
+        ))
+        return table_ref
```
vinceatbluelabs, line 402: I broke out this function so it could be easily unit tested with the various valid combinations of specifying project/dataset/table in the schema and table arguments.  I test drove this, and every line was added due to a corresponding test case.

test/test_sqlalchemy_bigquery.py
```diff
@@ -477,6 +483,38 @@ def test_get_columns(inspector, inspector_using_test_dataset):
...
+                             ('project.dataset', 'other_project.dataset.table', 'project'),
+                             ('project.dataset.something_else', 'table', 'project'),
+                             (None, 'project.dataset.table.something_else', 'project'),
+                         ])
+def test_invalid_table_reference(dialect, provided_schema_name,
+                                 provided_table_name, client_project):
+    with pytest.raises(ValueError):
+        dialect._table_reference(provided_schema_name,
+                                 provided_table_name,
+                                 client_project)
```
vinceatbluelabs, line 515: Above are the test cases!  Let me know if you think of any other scenarios we need to cover.

tswast:
Thanks for the contribution! It'll probably be another week or two before I have time to review this. Thanks for your patience.

tswast (approved):
Wonderful! Thanks so much for your contribution.

PR #47, opened 2020-02-10
Fix visit_label override to handle nested labels

JacobHayes:
Fixes #39 - passes the "test case" I posted there and a number of internal queries I've tested with. Will look into adding a formal test in the next few days.

```
Print:
SELECT sum(sum(`test`.`val`)) OVER () AS `anon_1`, sum(CASE WHEN %(param_1)s THEN `test`.`val` END) AS `sum_1`, sum(sum(CASE WHEN %(param_2)s THEN `test`.`val` END)) OVER () AS `anon_2`
FROM `test`

Compile:
SELECT sum(sum(`test`.`val`)) OVER () AS `anon_1`, sum(CASE WHEN %(param_1)s THEN `test`.`val` END) AS `sum_1`, sum(sum(CASE WHEN %(param_2)s THEN `test`.`val` END)) OVER () AS `anon_2`
FROM `test`

Literal compile:
SELECT sum(sum(`test`.`val`)) OVER () AS `anon_1`, sum(CASE WHEN true THEN `test`.`val` END) AS `sum_1`, sum(sum(CASE WHEN true THEN `test`.`val` END)) OVER () AS `anon_2`
FROM `test`

Result:
((6, 6, 6),)
```

I haven't noticed any regressions with expressions in the group by. That seems to be fully supported by BQ anyway, so I'm guessing the method I removed was just for nicer looking "group by" statements? I'll try to break this a bit more, but would love some other insight there.

PR #64, opened 2020-09-06
Compile array columns.

jmcarp:
Generate bigquery ddl for array columns and add tests.

cc @tswast @mxmzdlv

PR #67, opened 2020-09-26
Add support for array and struct literals

JacobHayes:
This allows building `array`s and `struct`s up in `sa.select` statements from other columns. The `array` is just ripped right from the postgres dialect with a custom `visit_array`. The `struct` have their own type and column expression. I didn't touch the existing `RECORD` type (which is just `JSON`) but instead made a custom `STRUCT` type that supports lookups to access subfields - not sure how to reconcile this with the existing `RECORD` type. Is it as easy as just changing `_type_map`'s value to `STRUCT` instead of `JSON`?

These have been useful for pivots (ie: array of struct("{X}" as name, {col} as value)).

~NOTE: due to struct fields requiring labels, things break when _not_ using `literal_binds` due to https://github.com/mxmzdlv/pybigquery/issues/39, but work fine after https://github.com/mxmzdlv/pybigquery/pull/47. I have an integration branch [here](https://github.com/JacobHayes/pybigquery/tree/integration) if it is useful in the meantime.~ #47 was merged

### DIFF
--- a/pybigquery/__init__.py
+++ b/pybigquery/__init__.py
@@ -1,1 +1,42 @@
 from sqlalchemy.dialects.postgresql import array
+from sqlalchemy.sql import expression, operators, sqltypes
+
+__all__ = ["array", "struct"]
+
+
+class STRUCT(sqltypes.Indexable, sqltypes.TypeEngine):
+    # NOTE: STRUCT names/types aren't currently supported.
+
+    __visit_name__ = "STRUCT"
+
+    class Comparator(sqltypes.Indexable.Comparator):
+        def _setup_getitem(self, index):
+            return operators.getitem, index, self.type
+
+    comparator_factory = Comparator
+
+
+class struct(expression.ClauseList, expression.ColumnElement):
+    """ Create a BigQuery struct literal from a collection of named expressions/clauses.
+    """
+    # NOTE: Struct subfields aren't currently propagated/validated.
+
+    __visit_name__ = "struct"
+
+    def __init__(self, clauses, field=None, **kw):
+        self.field = field
+        self.type = STRUCT()
+        super().__init__(*clauses, **kw)
+
+    def _bind_param(self, operator, obj, _assume_scalar=False, type_=None):
+        if operator is operators.getitem:
+            # TODO:
+            # - Validate field in clauses (or error if no clauses)
+            # - If the field is a sub-struct, return with all clauses, otherwise none.
+            return struct([], field=obj)
+
+    def self_group(self, against=None):
+        if not self.field and against in (operators.getitem,):
+            return expression.Grouping(self)
+        else:
+            return self
--- a/pybigquery/__init__.py
+++ b/pybigquery/__init__.py
@@ -1,0 +1,1 @@
+from sqlalchemy.dialects.postgresql import array
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -136,6 +136,19 @@ class BigQueryCompiler(SQLCompiler):
     def visit_array(self, element, **kw):
         return "[%s]" % self.visit_clauselist(element, **kw)
 
+    def visit_struct(self, element, within_columns_clause=True, **kw):
+        if element.field:
+            return self.preparer.quote_column(element.field)
+        kw["within_columns_clause"] = True
+        values = self.visit_clauselist(element, **kw)
+        return "struct(%s)" % values
+
+    def visit_getitem_binary(self, binary, operator, **kw):
+        return "%s.%s" % (
+            self.process(binary.left, **kw),
+            self.process(binary.right, **kw),
+        )
+
     def visit_select(self, *args, **kwargs):
         """
         Use labels for every column.
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -133,6 +133,9 @@ class BigQueryCompiler(SQLCompiler):
             kwargs['compile_kwargs'] = util.immutabledict({'include_table': False})
         super(BigQueryCompiler, self).__init__(dialect, statement, column_keys, inline, **kwargs)
 
+    def visit_array(self, element, **kw):
+        return "[%s]" % self.visit_clauselist(element, **kw)
+
     def visit_select(self, *args, **kwargs):
         """
         Use labels for every column.


JacobHayes:
Simple select example:
```python
import sqlalchemy as sa
from pybigquery import array, struct
from pybigquery.sqlalchemy_bigquery import BigQueryDialect

engine = sa.create_engine("bigquery://<your connstr here>")
engine.execute(
    """
    create table tmp.test_struct as (
        select 'a' as id, struct(1 as x__count, 2 as y__count, 3 as z__count) as dimensions
    )
    """
)

table = sa.Table("tmp.test_struct", sa.MetaData(bind=engine), autoload=True)
# This would obviously be cleaner if the RECORD built these natively.
dimensions = struct([], table.c.dimensions.name)
query = sa.select([table.c.id, dimensions["x__count"].label("x")])

print(query.compile(dialect=BigQueryDialect(), compile_kwargs={"literal_binds": True}))
```

Example doing a pivot:
```python
import sqlalchemy as sa
from pybigquery import array, struct
from pybigquery.sqlalchemy_bigquery import BigQueryDialect

engine = sa.create_engine("bigquery://<your connstr here>")
engine.execute(
    """
    create table tmp.test as (
        select 1 as x__count, 2 as y__count, 3 as z__count
    )
    """
)

table = sa.Table("tmp.test", sa.MetaData(bind=engine), autoload=True)
pivot_unnest = sa.func.unnest(
    array(
        [
            struct([sa.literal(dim).label("dimension"), col.label("count")])
            for dim, col in {
                "x": table.c.x__count,
                "y": table.c.y__count,
                "z": table.c.z__count,
            }.items()
        ]
    )
).alias("pivot")
pivot = struct([sa.column("dimension"), sa.column("count")], field=pivot_unnest.name)
query = sa.select(
    [pivot["dimension"].label("dimension"), pivot["count"].label("count")]
).select_from(table.join(pivot_unnest, sa.literal(True)))

print(query.compile(dialect=BigQueryDialect(), compile_kwargs={"literal_binds": True}))
```
It's a little bit awkward since they're not table objects (so no `.c` and instead do getitem style lookup). It seems like we may be able to mock that though with a little more `Comparator` work.

A few other things that'd be nice:
- `STRUCT` typing (ie: track name + types) - we should get this from both the BQ reflection and any dynamic `struct(...)` calls.
- validate field access against that metadata and determine return value datatype (including typed sub-structs)

tswast:
It might make sense to support ARRAY first, as there's already a data type for it in SQLAlchemy. https://docs.sqlalchemy.org/en/13/core/type_basics.html#sql-standard-and-multiple-vendor-types

Have you seen any discussion on adding STRUCT/RECORD types to SQLAlchemy for other backends? I'd want to make sure we do something that aligns with the direction of the rest of the community.

JacobHayes:
@tswast makes sense - the `ARRAY` datatype is handled with #64, correct? I don't see an `array` literal (`sa.select` instead of `sa.Table`) except [the postgres one](https://github.com/sqlalchemy/sqlalchemy/blob/640cd8a70f8a664b7834c5f74ec322fdea644043/lib/sqlalchemy/dialects/postgresql/array.py#L46), which might be because ["currently only the PostgreSQL backend has support for SQL arrays in SQLAlchemy"](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.ARRAY). I'm happy to pull out the `visit_array` and `array` import (or would you rather copy the [`class array`](https://github.com/sqlalchemy/sqlalchemy/blob/640cd8a70f8a664b7834c5f74ec322fdea644043/lib/sqlalchemy/dialects/postgresql/array.py#L46-L151) if you'd prefer.

Re a STRUCT/RECORD datatype - I didn't find anything on a cursory look at the sqla Google Group, they might be even more rare than `ARRAY` datatypes. 😁

Perhaps I can start a discussion in the Google Groups about these things.

jimfulton:
@JacobHayes wrt array literal, why not just `literal([2, 6])`

I guess this gets more exciting with structs, and especially arrays of structs.

JacobHayes:
@jimfulton oh yeah, hadn't thought of `literal`. That does seem to work for simple values, but not for column references:

```python
import sqlalchemy as sa
from pybigquery import array, struct
from pybigquery.sqlalchemy_bigquery import BigQueryDialect

engine = sa.create_engine("bigquery://")
engine.execute(
    """
    create or replace table tmp.test_struct as (
        select 'a' as id, struct(1 as x__count, 2 as y__count, 3 as z__count) as dimensions
    )
    """
)

table = sa.Table("tmp.test_struct", sa.MetaData(bind=engine), autoload=True)

print(list(engine.execute(sa.select([sa.literal(["a"])]))))
# [(['a'],)]
engine.execute(sa.select([sa.literal([table.c.id])]))
# ProgrammingError: (google.cloud.bigquery.dbapi.exceptions.ProgrammingError) Encountered unexpected first array element of parameter param_1, cannot determine array elements type.
# [SQL: SELECT %(param_1)s AS `anon_1`]
# [parameters: {'param_1': [Column('id', String(), table=<tmp.test_struct>)]}]
# (Background on this error at: http://sqlalche.me/e/13/f405)

print(list(engine.execute(sa.select([array(["a"])]))))
# [(['a'],)]
print(list(engine.execute(sa.select([array([table.c.id])]))))
# [(['a'],)]
```

jimfulton:
Thanks.  It's really helpful to have specific examples!

jimfulton:
FTR, IMO a big (surprising) part of the value proposition of SQLAlchemy is as an abstraction layer over disparate database engines.  For that reason, I'd like to enable as much as possible without introducing dialect-specific APIs.

JacobHayes:
Absolutely agree there! The array literal side at least matches postgres, but structs seem a bit more unique (though not so much for data warehouses). Perhaps upstream can be convinced to move the `array` constructor from the postgres dialect to common and (maybe harder) add a `struct`?

Additionally, a proper `STRUCT` column type would be great so we can start generating DDL for more complex tables (JSON won't do as it doesn't define the subfields).

jimfulton:
@JacobHayes please checkout https://github.com/googleapis/python-bigquery-sqlalchemy/pull/318/files#diff-ee8781d1be11dbad05f4936052659eed168b4ebeaef166b9e2684f74293e30f8R94-R183

and let me know if I satisfied what you were trying to do.

I didn't really follow your pivot example, but I'm hoping I covered it in [`test_unnest_and_struct_access_233`](https://github.com/googleapis/python-bigquery-sqlalchemy/pull/318/files#diff-ee8781d1be11dbad05f4936052659eed168b4ebeaef166b9e2684f74293e30f8R127-R183).

jimfulton:
Superseded by #318

PR #70, opened 2020-10-01
[WIP] Register Alembic dialect. Fix README

mvoitko:
I started adding Alembic support as mentioned [here](https://github.com/sqlalchemy/alembic/issues/738), but I need help to define table altering instructions in BigQuery.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,10 @@
 /.vscode/
 *.pyc
 .cache/
+.idea/
 *.iml
 /pybigquery.egg-info
 /dist
 /build
 env
+venv/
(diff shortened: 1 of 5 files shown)


tswast:
Just weeks after you sent this PR, some limited ALTER TABLE support was added. :-) https://cloud.google.com/bigquery/docs/release-notes?hl=en#October_14_2020

pedro-iatzky:
Can we have this feature merged?
I think it is very powerful even if we can only execute SQL statements with the op.execute() method?

jimfulton:
Closing in favor of #183.  

Some notes:

- I would have started from this PR, but there were so many changes in the underlying code, it was easier to start over.
- The registration code was pretty much all that was needed.
- This package shouldn't depend on alembic, but leverage it if it's available.
- #183 includes documentation and a system test showing what works, which is quite a lot.

Issue #83, opened 2021-03-17
Problem with SQLAlchemy 1.4.0


kromash:
Hello,

New version of SQLAlchemy (1.4.0) was released 15.03.2020 and since then I got following error:
```
 File "/layers/google.python.pip/pip/lib/python3.8/site-packages/promise/promise.py", line 489, in _resolve_from_executor
    executor(resolve, reject)
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/promise/promise.py", line 756, in executor
    return resolve(f(*args, **kwargs))
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/graphql/execution/middleware.py", line 75, in make_it_promise
    return next(*args, **kwargs)
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/graphene_sqlalchemy/fields.py", line 78, in connection_resolver
    return on_resolve(resolved)
  File "src/adtrender_api/connections.py", line 52, in resolve_connection
    return super().resolve_connection(connection_type, model, info, args, resolved)
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/graphene_sqlalchemy/fields.py", line 53, in resolve_connection
    _len = resolved.count()
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3009, in count
    return self._from_self(col).scalar()
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 2753, in scalar
    ret = self.one()
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 2730, in one
    return self._iter().one()
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 2768, in _iter
    result = self.session.execute(
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 1653, in execute
    result = conn._execute_20(statement, params or {}, execution_options)
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1520, in _execute_20
    return meth(self, args_10style, kwargs_10style, execution_options)
  File "/layers/google.python.pip/pip/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 313, in _execute_on_connection
[...]

PR #84, opened 2021-03-19
feat: add support for SQLAlchemy 1.4

alonme:
Still a draft, need to make sure all this changes will not break users in older versions of SQLA


alonme:
closes #83

tswast:
Thanks! Let me know once this is ready for review.

alonme:
@tswast - Almost ready, 
I'm debating about having SQLA 1.4 a dev requirement (for tests)
and about 1 issue of backward compatibility, which might be an issue only if someone actually passed `inline` to `BigQueryCompiler`

tswast:
> I'm debating about having SQLA 1.4 a dev requirement (for tests)

Ideally we'd have some test sessions for our minimum requirements as well as the latest supported. I've created https://github.com/mxmzdlv/pybigquery/issues/85 to track this.

tswast:
In the meantime, I'm okay adding a dev requirement.

tswast:
@alonme Apologies, but I just "blackened" all the source in this repo. It will most likely cause some conflicts.

In good news, we do now have "constraints" files in the `testing/` directory which allow us to test against specific package versions for each of our test sessions. Plain ol' pytest should still work, but to use that multi-version testing, we can run with `nox` http://nox.thea.codes/

jimfulton:
See: #177

PR #129, opened 2021-04-20
fix: Select expressions no-longer force use of labels

### LINKED ISSUE #78: don't enforce "use_labels = True"
`BigQueryCompiler.visit_select` sets `args[0].use_labels = True`
in my project for example this is unneeded and changes all column names.
Please either:
1. provider a way to ignore that assignment
2. provide a way to "translate" back column name as label to its original name

image

jimfulton:

Fixes #78  🦕

### DIFF
--- a/pybigquery/sqlalchemy_bigquery.py
+++ b/pybigquery/sqlalchemy_bigquery.py
@@ -165,15 +165,6 @@ class BigQueryCompiler(SQLCompiler):
             dialect, statement, column_keys, inline, **kwargs
         )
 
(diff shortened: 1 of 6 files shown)


tswast (approved):
Thanks for adding unit tests!

Interesting to use SQLite for unit testing, but I can roll with that.

Long-term we might want to do something with Zeta SQL (name of the BigQuery/Spanner open source implementation), but I think it's a long ways away before we get something usable like a local emulator out of that project.

PR #166, opened 2021-05-12
chore: Make it practical to make simultaneous compliance test runs.

### LINKED ISSUE #165: Allow multiple compliance test runs
It should be possible to run tests locally while they're running in CI.  To do that, we need to use different test schemas.  

I ended up changing compliance tests to use randomly-named temporary schemas.

jimfulton:

Fixes #165 🦕

This changes SQLAlchemy-provided compliance tests to run in temporary datasets, so multiple test runs can happen simultaneously.

Unfortunately, this requires disabling the tests that verify that you can refer to tables with dataset/schema names `mydataset.mytable` be disabled by default, as the schema used by these tests is hard-coded.  These tests can still be run manually using `--requirements pybigquery.requirements:WithSchemas`.  The system tests exercise this functionality.

BTW, the system tests need the same treatment.  That will be in a separate PR.


plamut (approved):
LGTM

Issue #252, opened 2021-08-17
Query compilation breaks using the literal_binds kwarg with an IN operator in the Query on SQLAlchemy 1.4


dylanpbacon:
#### Environment details

  - OS type and version: MacOS Big Sur 11.5.1
  - Python version: 3.7.4
  - pip version: 20.3.3
  - `sqlalchemy-bigquery` version: 0.10.0
  - `sqlalchemy` version: 1.4.22

#### Steps to reproduce

  1.  Create a SQLAlchemy query using an `in_` or `notin_` parameter.
  2. Attempt to compile this query to text using the `compile()` method with `{"literal_binds": True}` in the kwargs. This worked when running with SQLAlchemy version 1.3 and pybigquery version 0.5.0, but does not work with SA 1.4 and pybigquery 0.10.0.

#### Code example

```python
from contextlib import contextmanager

from sqlalchemy import MetaData, Table, create_engine, select

engine = create_engine("bigquery://<your bigquery connection here>")
# Postgres connection, for comparison of proper literal binding and compilation
#engine = create_engine("postgresql://<your postgres connection here>")


@contextmanager
def test_query():
    engine.execute(
        """
        create table test as (
            select 1 as val
            union all select 2
            union all select 3
        )
    """
    )
    val = Table("test", MetaData(bind=engine), autoload=True).c.val
    try:
        yield select(
            [val]
        ).where(val.in_([2]))
    finally:
        engine.execute("drop table test")


with test_query() as q:
    print("Print:")
    print(q)

    print("\nCompile:")
    print(q.compile(engine))

    print("\nLiteral compile:")
    print(q.compile(engine, compile_kwargs={"literal_binds": True}))

    print("\nResult:")
    print(tuple(engine.execute(q)))

```

#### Stack trace
```
Print:
SELECT `test`.`val` 
FROM `test` 
WHERE `test`.`val` IN UNNEST([ [POSTCOMPILE_val_1]:INT64 ])

Compile:
SELECT `test`.`val` 
FROM `test` 
WHERE `test`.`val` IN UNNEST([ [POSTCOMPILE_val_1]:INT64 ])

Literal compile:
Traceback (most recent call last):
  File "/Users/admin/Desktop/test_bq_in.py", line 37, in <module>
    print(q.compile(engine, compile_kwargs={"literal_binds": True}))
[...]

PR #253, opened 2021-08-17
feat: Handle passing of arrays to in statements more efficiently in SQLAlchemy 1.4 and higher

### LINKED ISSUE #194: Support the passing of arrays as single arguments to BigQuery
Hey guys! thanks for the awesome plugin!

BigQuery only supports 10K parameters to be passed in a single query. while this is enough for the most part, sometimes that could cause real issues, for example:
```python
book_ids = [2,5,22,3123,124123,...]  # a list of 23K book ids
session.query(Book.name).filter(Book.book_id.in_(book_ids)).all()
# this returns a BadRequest (400) for having too many parameters (each id is translated to a parameter.
```
Fortunately, BQ supports passing an array as a single parameter, but SQLAlchemy+pybigquery doesn't take advantage of that. I have make a little workaround to fulfill my own needs, I guess this could be used in the pybigquery implementation of `Column.in_`
```python
import random
from sqlalchemy import bindparam, text
def _workaround_in(schema, column, iterable):
    bind_name = column.key + str(random.randint(0, 1_000_000))
    bound = bindparam(bind_name, iterable)
    return text(
        f"`{schema.__tablename__}`.`{column.key}` in UNNEST(:{bind_name})"
    ).bindparams(bound)

# usage:
book_ids = [2,5,22,3123,124123,...]  # a list of 23K book ids
session.query(Book.name).filter(_workaround_in(Book, Book.book_id, book_ids)).all()
# this will work!
```
I did try to split the statement in the query, but it didn't help.
the only other thing that worked was making the entire query with text.

jimfulton:

Fixes #194 🦕


jimfulton:
**For reviewers:**

When dealing with an in expression, like `sqlalchemy.literal(-1).in_(list(range(99999)))` (see `test_huge_in`), SQLAlchemy wants to turn each element of the passed list into a separate parameter, which is inefficient and breaks for large lists.

BigQuery lets you pass array parameters, which is much better for this case.

For SQLAlchemy 1.4, it's fairly easy to disable this behavior. I beat my head against earlier versions and gave up. :)  The only trick is to make sure we call `UNNEST` on the parameter in SQL.  We have to do that in the non-array case too, although it's more complicated.

Finally, `in`-related tests:

- Have to expect different SQL for SQLAlchemy 1.4 and earlier versions.
- For unit tests, because we're using sqlite, we don't get correct results, because sqlite doesn't support arrays, so we only check the SQL generated.

jimfulton:
I added a closely related change to fix #252.  As it happens, if "literal binds" are requested, the work we do in `visit_bindparam` is unnecessary and we can return the `super` result. 

In making that change, I realized that the change I made earlier in the PR could be simplified, making the critical part of the diff smaller.

tests/unit/test_select.py
```diff
@@ -239,63 +228,74 @@ def test_select_in_lit_13(faux_conn):
...
-    )
-    assert isin
-    assert _normalize_in_params(*faux_conn.test_data["execute"][-1]) == (
-        "SELECT %(p_0:INT64)s IN "
-        "UNNEST([ %(p_1:INT64)s, %(p_2:INT64)s, %(p_3:INT64)s ]) AS `anon_1`",
-        {"p_1": 1, "p_2": 2, "p_3": 3, "p_0": 1},
+def test_select_in_lit(faux_conn, last_query):
+    faux_conn.execute(sqlalchemy.select([sqlalchemy.literal(1).in_([1, 2, 3])]))
+    last_query(
+        "SELECT %(param_1:INT64)s IN UNNEST(%(param_2:INT64)s) AS `anon_1`",
```
tswast, line 234: Should the type of `param_2` be `ARRAY<INT64>`?

jimfulton: No, because the BQ DB API does special handling of arrays.

It sees that we have a scalar type of INT64 and that we have a sequence, and then creates a `ArrayQueryParameter`.

It happens that since I added struct support, passing `ARRAY<INT64>` would _probably_ work (because I have to handle structs of arrays).  But just usng `INT64` works too.

tswast (approved):

PR #279, opened 2021-08-18
chore: Update many dependencies

jimfulton:

Closes #276 
Closes #270 
Closes #269 
Closes #268 
Closes #264 
🦕


tswast (approved):

samples/snippets/requirements.txt
```diff
@@ -8,7 +8,7 @@ click==8.0.1
 click-plugins==1.1.1
 cligj==0.7.2
 contextvars==2.4
-dataclasses==0.6
+dataclasses==0.6; python_version < '3.7'
```
tswast, line 11: Presumably we can use dataclasses 0.8 here, right?

jimfulton: We could, but since it's included in Python, why bother?

That's what I took from your comment on #264:

> "No distribution found" -- Maybe this needs to be excluded from versions where the backport is not needed?

PR #290, opened 2021-08-20
fix: unnest failed in some cases (with table references failed when there were no other references to refrenced tables in a query)

### LINKED ISSUE #283: `unnest` isn't handled correctly
Given a table, `foo`, with schema:
`[SchemaField('objects', 'RECORD', 'REPEATED', None, (SchemaField('object_id', 'STRING', 'NULLABLE', None, (), ()),), ())]`

```
>>> foo = Table('foo', metadata, autoload_with=engine)
>>> print(select([func.unnest(foo.c.objects).alias('foo_objects')]).compile(engine))
SELECT `foo_objects`.`unnest_1` 
FROM unnest(`foo`.`objects`) AS `foo_objects`
```

BigQuery needs this select to be spelled:

```
SELECT foo_objects
FROM `foo`, UNNEST(objects) as foo_objects
```

(The examples above don't show a dataset, because it's set as a default.)

jimfulton:

Fixes #283  🦕


jimfulton:
Reviewer:

This PR is explained pretty well, I think, by the comment here:
https://github.com/googleapis/python-bigquery-sqlalchemy/pull/290/files#diff-efe1987a73ae5b54aa817dc8c70541025df89d9d08601c8ab33fa84ad8c09de7R254-R272

Mostly unrelated, I added the missing `ARRAY` type to the list of types, to facilitate a test.  I took the opportunity to sort the types, which contributes a quite a few changed lines, but the change is trivial.

sqlalchemy_bigquery/base.py
```diff
@@ -246,6 +250,56 @@ def visit_insert(self, insert_stmt, asfrom=False, **kw):
...
+        #
+        #   SELECT foo_objects
+        #   FROM `foo`, UNNEST(`foo`.`objects`) as foo_objects
+        #
+        # This is tricky because:
+        # 1. We have to find the table references.
+        # 2. We can't know practically if there's already a FROM for a table.
+        #
+        # We leverage visit_column to find a table reference.  Whenever we find
+        # one, we create an alias for it, so as not to conlfict with an existing
```
tswast, line 280:
```suggestion
        # one, we create an alias for it, so as not to conflict with an existing
```

sqlalchemy_bigquery/base.py
```diff
@@ -246,6 +250,56 @@ def visit_insert(self, insert_stmt, asfrom=False, **kw):
...
+        # reference if one is present.
+        #
+        # This requires communicating between this function and visit_column.
+        # We do this by sticking a dictionary in the keyword arguments.
```
tswast, line 284: Is it a documented feature of SQLAlchemy that it'll always use `**kwargs` to pass through between these functions? If so, can we link to such a document?

If not, I'm a bit wary of this method of communication, though I don't have any better ideas.

tests/system/test_sqlalchemy_bigquery.py
```diff
@@ -34,6 +34,7 @@
 import datetime
 import decimal
 
+sqlalchemy_version_info = tuple(map(int, sqlalchemy.__version__.split(".")))
```
tswast, line 37: This works, though I've been trying to use https://packaging.pypa.io/en/latest/version.html which is the canonical version parser. We already have it pulled in via a transitive dependency through `setuptools`, I believe.

sqlalchemy_bigquery/base.py
```diff
@@ -114,39 +116,41 @@ def format_label(self, label, name=None):
...
     "TIME": types.TIME,
-    "RECORD": types.JSON,
-    "NUMERIC": types.Numeric,
-    "BIGNUMERIC": types.Numeric,
 }
 
 # By convention, dialect-provided types are spelled with all upper case.
-STRING = _type_map["STRING"]
-BOOL = _type_map["BOOL"]
+ARRAY = _type_map["ARRAY"]
```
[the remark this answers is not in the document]
jimfulton: It's not a new feature.  It was just missing from the types list.  

SQLAlchemy has basic array support that works with BigQuery (thanks to some previous effort on our part).

sqlalchemy_bigquery/base.py
jimfulton: That's a good question.  I've seen it so much in the code that I've just taken it for granted.  

I asked here: https://github.com/sqlalchemy/sqlalchemy/discussions/6931.

Often options are explicit keyword arguments that become `**kwargs`, usually by necessity because arguments get passed through many intermediate layers that don't care about them.

In the case, an alternative would be to add a private/`__` attribute to the alias element, which is passed by SQLAlchemy to `visit_column` as the   `'enclosing_alias'` key in `kwargs`.  :)  However, relying on our own key seemed safer to me.

I'm curious to see what response I get in the discussion.

tests/system/test_sqlalchemy_bigquery.py
jimfulton: Nice. I'll switch to that.

WRT how we get it, IMO, anything we import should be a direct dependency.

sqlalchemy_bigquery/base.py
tswast: > So that said if you are using your own values in the **kw passed through compiler.process() and all the visit_ methods, that's fine, might be a good idea to qualify the names you use so they don't conflict with future names.

Sounds like you're doing the right thing here.

tests/system/test_sqlalchemy_bigquery.py
tswast: Sounds good. I'm fine explicitly including it in our dependencies / test dependencies.

jimfulton: Done

jimfulton:
@tswast are you cool with this? :)

jimfulton:
I have to more small PRs coming that will build on this.

tswast (approved):

PR #361, opened 2021-10-26
fix: avoid creating aliases for already-known tables

### LINKED ISSUE #353: Unnest function in BigQuery SQLAlchemy plugin joins twice when selecting other columns
From a customer support request:

Using the UNNEST function generates a CROSS JOIN with a table to itself.
A minimal example is included below. The following Python code:

```
from sqlalchemy import create_engine, MetaData, select, func as F, Table, Column, Integer, ARRAY
metadata = MetaData()
table = Table(
  "table1",
  metadata,
  Column("foo", Integer),
  Column("bar", ARRAY(Integer))
)
q = select(table.c.foo).where(F.unnest(table.c.bar).column_valued() == 1)
print(q.compile(engine))
```

Generates this query:

```
SELECT `table1`.`foo` 
FROM `table1`, `table1` `table1_1`, unnest(`table1_1`.`bar`) AS `anon_1` 
WHERE `anon_1` = %(param_1:INT64)s
```

I expect it to generate this query instead (note the different FROM clause):

```
SELECT `table1`.`foo` 
FROM `table1`,  unnest(`table1`.`bar`) AS `anon_1` 
WHERE `anon_1` = %(param_1:INT64)s
```

tseaver:
Closes #353.

### DIFF
--- a/sqlalchemy_bigquery/base.py
+++ b/sqlalchemy_bigquery/base.py
@@ -50,6 +50,7 @@ from sqlalchemy.sql.sqltypes import Integer, String, NullType, Numeric
 from sqlalchemy.engine.default import DefaultDialect, DefaultExecutionContext
 from sqlalchemy.engine.base import Engine
 from sqlalchemy.sql.schema import Column
+from sqlalchemy.sql.schema import Table
 from sqlalchemy.sql import elements, selectable
 import re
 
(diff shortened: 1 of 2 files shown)


tswast (approved):
Nice!

PR #410, opened 2022-02-04
feat: Allow base64 encoded credentials in URI

### LINKED ISSUE #409: Allow base64 encoded credentials in URL

I'm using [Intake](https://github.com/intake/intake) to build a catalog server where the server stores the credentials and does not require our users to have the credentials on their clients. When using the Intake server with `direct_access: allow` the current scenario creates the SQLAlchemy engine by forwarding the connection string from the Intake server to the client. The alternative is `direct_access: forbid` which means that the Intake server connects to the database with its stored credentials json file and then forwards the data to the client. This creates a memory bottleneck in the Intake server. The `direct_access: allow` method forwards the connection string to the client to make the connection by passing `credentials_path=/path/on/server/creds.json` means that the client cannot authenticate.

I'd like to have the full credentials inlined into the URL to enable our Intake server with `direct_access: allow`. While in a general case this may be less than ideal since this catalog server is forwarding secret information to the client, we deem it be acceptable in our case.

Many other SQLAlchemy drivers accept username/password credentials in their URLs so this does not feel any worse to me.

I have prepared a PR that I will submit.

AlbertDeFusco:
Fixes #409 🦕 To enable credential information to be included in the connection URL, for cases where you don't have a credentials file locally on the client, I propose the `credentials_base64` parameter. It requires that the user have encoded their credentials JSON file using a number of techniques like `base64`, or `openssl base64`, or `python -m base64`, or www.base64encode.org.

I have used nox to run unit and system tests for Python 3.6 - 3.9. I'm tracking down a separate issue with my computer for why 3.10 tests did not run.

### DIFF
--- a/tests/unit/test_helpers.py
+++ b/tests/unit/test_helpers.py
@@ -115,7 +115,7 @@ def test_create_bigquery_client_with_credentials_base64(monkeypatch, module_unde
     monkeypatch.setattr(service_account, "Credentials", mock_service_account)
 
     credentials_info = (
-        {"type": "service_account", "project_id": "service-account-project",},
+        {"type": "service_account", "project_id": "service-account-project"},
     )
 
     credentials_base64 = base64.b64encode(json.dumps(credentials_info).encode())
@@ -141,7 +141,7 @@ def test_create_bigquery_client_with_credentials_base64_respects_project(
     monkeypatch.setattr(service_account, "Credentials", mock_service_account)
 
     credentials_info = (
-        {"type": "service_account", "project_id": "service-account-project",},
+        {"type": "service_account", "project_id": "service-account-project"},
     )
 
     credentials_base64 = base64.b64encode(json.dumps(credentials_info).encode())
--- a/tests/unit/test_helpers.py
+++ b/tests/unit/test_helpers.py
@@ -4,12 +4,14 @@
 # license that can be found in the LICENSE file or at
 # https://opensource.org/licenses/MIT.
 
+import base64
+import json
 from unittest import mock
 
 import google.auth
 import google.auth.credentials
-from google.oauth2 import service_account
 import pytest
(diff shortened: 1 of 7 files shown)


tswast (approved):

tests/unit/test_helpers.py
```diff
@@ -105,6 +107,52 @@ def test_create_bigquery_client_with_credentials_info_respects_project(
...
 
+def test_create_bigquery_client_with_credentials_base64(monkeypatch, module_under_test):
+    mock_service_account = mock.create_autospec(service_account.Credentials)
+    mock_service_account.from_service_account_info.return_value = AnonymousCredentialsWithProject(
+        "service-account-project"
+    )
+    monkeypatch.setattr(service_account, "Credentials", mock_service_account)
+
+    credentials_info = (
+        {"type": "service_account", "project_id": "service-account-project",},
```
tswast, line 118:
```suggestion
        {"type": "service_account", "project_id": "service-account-project"},
```

tests/unit/test_helpers.py
```diff
@@ -105,6 +107,52 @@ def test_create_bigquery_client_with_credentials_info_respects_project(
...
+    https://github.com/googleapis/python-bigquery-sqlalchemy/issues/48
+    """
+    mock_service_account = mock.create_autospec(service_account.Credentials)
+    mock_service_account.from_service_account_info.return_value = AnonymousCredentialsWithProject(
+        "service-account-project"
+    )
+    monkeypatch.setattr(service_account, "Credentials", mock_service_account)
+
+    credentials_info = (
+        {"type": "service_account", "project_id": "service-account-project",},
```
tswast, line 144:
```suggestion
        {"type": "service_account", "project_id": "service-account-project"},
```

tswast (approved):

PR #401, opened 2022-01-18
test: add prerelease session to test with latest dependencies, separate compliance tests

tswast:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-sqlalchemy/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Towards https://github.com/googleapis/python-bigquery/issues/1021 "Test connectors, such as pandas-gbq and sqlalchemy-bigquery with pre-release, make updates if needed."

Towards https://github.com/googleapis/python-bigquery-sqlalchemy/issues/413

Internal CL 422873148 to enable these test sessions in Kokoro.

PR #414, opened 2022-02-17
deps: require google-cloud-bigquery-storage to avoid performance warning

tswast:
The wrapped DB-API uses google-cloud-bigquery-storage to download query
results by default. It raises a warning and falls back to the REST API
if google-cloud-bigquery-storage is not installed.

The google-cloud-bigquery version 3.x library is tested in #401 and it does not include any breaking changes that should affect the SQLAlchemy connector.

deps: allow google-cloud-bigquery version 3.x

- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)


setup.py
```diff
@@ -84,7 +84,8 @@ def readme():
         # Until this issue is closed
         # https://github.com/googleapis/google-cloud-python/issues/10566
         "google-auth>=1.25.0,<3.0.0dev",  # Work around pip wack.
-        "google-cloud-bigquery>=2.25.2,<3.0.0dev",
+        "google-cloud-bigquery>=2.25.2,<4.0.0dev",
+        "google-cloud-bigquery-storage>=2.0.0,<3.0.0dev",
```
tswast, line 88: Need pyarrow as well. Probably a good thing we're adding this, because I totally missed that.

parthea (approved):

PR #431, opened 2022-03-18
fix: use faux_conn rather than engine in unit tests

### LINKED ISSUE #430: unit tests require google application credentials
The tests in tests/unit/test_select and test__struct require application credentials to run. Without credentials, running the suite results in errors like:

```
raise exceptions.DefaultCredentialsError(_HELP_MESSAGE)
google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see [https://cloud.google.com/docs/authenticati\](https://cloud.google.com/docs/authenticati%5C)
on/getting-started
```

The credential requirement prevents them from being pure unit tests and running in any environment.

waltaskew:
Remove the engine fixture from unit tests and use faux_conn instead. Creating
an engine requires credentials, which prevents these tests from being pure unit
tests and running in any environment.

Fixes #430 🦕

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -43,11 +43,6 @@ sqlalchemy_before_1_4 = pytest.mark.skipif(
 
 
 @pytest.fixture()
-def engine():
-    return sqlalchemy.create_engine("bigquery://myproject/mydataset")
-
-
-@pytest.fixture()
 def faux_conn():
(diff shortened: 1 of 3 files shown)


tswast (approved):
Thanks!

PR #433, opened 2022-03-18
fix: use explicit rather than implicit relative imports

### LINKED ISSUE #432: use explicit rather than implicit relative imports
Some tests make use of implicit relative imports (e.g. `import fauxdbi`) which don't work in all environments. Using explicit relative imports (e.g. `from . import fauxdbi`) makes it easier to use the package and run the test suite in more environments.

waltaskew:
use explicit rather than implicit relative imports

Fixes #432 🦕


tswast (approved):

tswast:
Thanks for the contribution!

PR #446, opened 2022-04-03
chore: use pytest-flake8==1.1.0 in dev_requirements.txt

### LINKED ISSUE #445: dev_requirements file has incompatible requirements
#### Summary
When trying to install development requirements on a fresh repo, an error is thrown from pip. This is because we've upgraded the dev_requirements.txt file to include a dependency on pytest-flake8 1.1.1, but that library is only compatible with pytest 7 and above. 

#### Environment details

  - Windows 11, WSL2 running Debian GNU/Linux 11 (bullseye)
  - Python version: 3.9.10
  - pip version: 22.0.3

#### Steps to reproduce

  1. Clone down the repository (or a fork) and check out the main branch (currently https://github.com/googleapis/python-bigquery-sqlalchemy/commit/f545d4ca661b77e309f5e95e9596a9d687bdd0a9)
  2. Create a virtualenv. Python version shouldn't matter, but this is definitely reproducible on 3.9.
  3. Try to install the dev requirements by doing `pip install -r dev_requirements.txt`.

#### Stack trace

```shell
INFO: pip is looking at multiple versions of pytest to determine which version is compatible with other requirements. This could take a while.
Collecting pytest==6.2.5
  Using cached pytest-6.2.5-py3-none-any.whl (280 kB)
INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of future to determine which version is compatible with other requirements. This could take a while.
Collecting future==0.18.2
  Using cached future-0.18.2-py3-none-any.whl
ERROR: Cannot install -r dev_requirements.txt (line 5) and pytest==6.2.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pytest==6.2.5
    pytest-flake8 1.1.1 depends on pytest>=7.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

youcandanch:
#### Summary

`dev_requirements.txt` was recently changed to upgrade the dependency on pytest-flake8 from 1.1.0 to 1.1.1. Unfortunately, pytest-flake8 requires pytest 7 or above, and the current pinned dependency for pytest is 6.2.5. I tried moving up to pytest 7 but it seems like there's a decent amount of work going into that, so in order to fix main, this PR moves the pytest-flake8 pinned version back to 1.1.0.

Fixes #445.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -2,5 +2,5 @@ sqlalchemy>=1.1.9
 google-cloud-bigquery>=1.6.0
 future==0.18.2
 pytest==6.2.5
-pytest-flake8==1.1.1
+pytest-flake8==1.1.0  # versions 1.1.1 and above require pytest 7
 pytz==2022.1


parthea (approved):

parthea:
Thanks @youcandanch !

PR #459, opened 2022-05-11
feat: GCP access token support

### LINKED ISSUE #42: Use OAuth to authenticate against BigQuery
Hi,

Is it possible to use OAuth to authentitcate against BigQuery, using end user identity ?

Thanks.
Yann

Freakachoo:
Add support for using a temporary GCP Access Token for authenticating.

# Changelog
- added property `credentials_access_token` for `create_engine`
- added unit tests
- updated Readme

Fixes #42  🦕

### DIFF
--- a/sqlalchemy_bigquery/_helpers.py
+++ b/sqlalchemy_bigquery/_helpers.py
@@ -10,7 +10,7 @@ import re
 from google.api_core import client_info
 import google.auth
 from google.cloud import bigquery
-from google.oauth2 import service_account
+from google.oauth2 import service_account, credentials as oauth_credentials
 import sqlalchemy
 import base64
 import json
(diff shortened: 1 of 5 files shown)


sqlalchemy_bigquery/_helpers.py
```diff
@@ -54,6 +55,9 @@ def create_bigquery_client(
         )
         credentials = credentials.with_scopes(SCOPES)
         default_project = credentials.project_id
+    elif credentials_access_token:
+        credentials = oauth_credentials.Credentials(credentials_access_token)
+        _, default_project = google.auth.default(scopes=SCOPES)
```
tswast, line 60: This is potentially a slow operation. I we might consider only doing this if `project_id` is `None`. Or maybe add a `if default_project is None` to the `if project_id is None` if statement below?

tswast (commented):
Looks good! Just one potential optimization.

tswast:
An optional `refresh_token` might also be desirable, but not required nor applicable to all use cases.

meredithslota:
This PR is pretty stale, opting to close since there haven't been responses to the code review comments. Thanks!

PR #470, opened 2022-07-11
Remove pyarrow from setup.py

bnaul:
#414 added `pyarrow` to the `setup.py` as part of adding `google-bigquery-storage` to `install_requires`. I understand the motivation for adding `google-bigquery-storage`, but since that has its own allowed `pyarrow` version range, it seems simpler to just rely on that rather than specifying yet another `pyarrow` range here which can get (/ currently is) out of sync with that library.

@tswast what do you think?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,6 @@ setup(
         "google-auth>=1.25.0,<3.0.0dev",  # Work around pip wack.
         "google-cloud-bigquery>=2.25.2,<4.0.0dev",
         "google-cloud-bigquery-storage>=2.0.0,<3.0.0dev",
-        "pyarrow>=3.0.0,<7.0dev",
         # Temporarily set maximimum sqlalchemy to a known-working version while
         # we debug failing compliance tests. See:
         # https://github.com/googleapis/python-bigquery-sqlalchemy/issues/386


tswast:
Related: https://github.com/googleapis/python-bigquery/issues/1196 is an open issue for a request to make pyarrow optional again in `google-cloud-bigquery`

tswast:
> Does that make sense or am I missing something about how this library relies on pyarrow?

The DB-API package that this depends on can optionally rely on the BigQuery Storage API, which requires pyarrow to deserialize the results. This can be much faster for downloading large results, but won't make much of a difference for smaller result sizes.

tswast:
If we were to drop pyarrow, we should also drop bqstorage. Ideally, we'd include these as "extras" and make sure we have tests where these packages are not installed & installed.

chalmerlowe:
> Related: [googleapis/python-bigquery#1196](https://github.com/googleapis/python-bigquery/issues/1196) is an open issue for a request to make pyarrow optional again in `google-cloud-bigquery`

1196 got closed via this PR:

https://github.com/googleapis/python-bigquery/pull/1282/files

This makes `pyarrow` and `google-cloud-bigquery-storage` optional

chalmerlowe:
@bnaul closing this PR based on some work related to making both `pyarrow` and `bqstorage` optional in PR #847.
Thanks so much for issuing this PR and providing impetus to improving `python-bigquery-sqlalchemy`.

PR #474, opened 2022-07-21
feat: Allow Users to Supply Their Own BigQuery Client

### LINKED ISSUE #464: Ability to Create Engine From an Existing BigQuery Client
In situations where a user already has a `google.cloud.bigquery.Client` object, it'd be helpful to be able to create an sqlalchemy engine either using that already existing BigQuery client object or the same credentials used by that client. 

I'm poking around and one option would be something like

```python
bq_client: bigquery.Client
engine = sqlalchemy.engine.create_engine(
    'bigquery://project/?credentials_user_info_json=%s' % bq_client._credentials.to_json())
```
to create a client with the same credentials. This is unfortunately accessing a private `_credentials` attribute on the client. We might look to adding a public interface for this on the client so callers don't need to access `_credentials`.

I was also looking at something like

```python
bq_client: bigquery.Client
engine = sqlalchemy.engine.create_engine(
  'bigquery://project', connect_args={'client': bq_client})
```
using the [connect_args kwarg](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine.params.connect_args)

The problem is that the engine will still try to create the client itself even if it will end up being replaced by the one from `connnect_args`. That attempt to create the client by the engine may fail (we're assuming the big query client is created by some machinery that's opaque to the user here.) I'm not seeing a clean way to avoid this.

Any other suggestions? I'm more than happy to work on this if it seems like a reasonable thing to support!

waltaskew:
- Add a flag 'user_supplied_client' which prevents the library from attempting
  to create a BigQuery client.
- Document the use of `connect_args` for suppling their own BigQuery client
  to the dbapi

Fixes #464 🦕


DimaK415:
Please implement this!  Will really help with Airflow use cases

parthea:
@chalmerlowe Please could you take a look?

chalmerlowe:
I will look at this and provide some feedback.

README.rst
```diff
@@ -234,6 +234,25 @@ To create the base64 encoded string you can use the command line tool ``base64``
...
+
+.. code-block:: python
+
+    from google.cloud import bigquery
+
+    custom_bq_client = bigquery.Client(...)
+
+    engine = create_engine(
+        'bigquery://some-project/some-dataset?user_supplied_client=True',
+	connect_args={'client': bq_client},
```
DimaK415, line 252: `bq_client` should be `custom_bq_client`

DimaK415:
Just a heads up.  I've tested the functionality of this PR and it seems to work.  No issues found so far.

README.rst
waltaskew: fixed!

DimaK415:
Anything I can help with to get this moving?

chalmerlowe (approved):
LGTM

chalmerlowe:
@waltaskew @DimaK415 

Thank you for you assistance in getting this over the finish line.

PR #479, opened 2022-08-10
fix(deps): allow pyarrow version 7+

### LINKED ISSUE #478: Support recent PyArrow versions

This package's [pyarrow requirement](https://github.com/googleapis/python-bigquery-sqlalchemy/blob/main/setup.py#L88) is effectively pinned to 6.0.1, which was released almost 9 months ago. This is preventing us from upgrading to new versions of other packages requiring more recent versions of pyarrow. e.g. the latest snowflake-connector-python requires pyarrow 8.

Bump your pyarrow requirement to support as many and as recent versions as possible.

Right now we're waiting to upgrade packages with conflicting pyarrow dependencies, but we're considering pinning pyarrow to a newer version and testing to see if this package still works (for our implementation, at least)

willsthompson:
   - One exception: I couldn't test 3.7 because PyArrow doesn't provide a macos/arm64 wheel for 3.7, and I couldn't get the build working on my machine. PyArrow does [claim to be compatible with 3.7](https://arrow.apache.org/docs/python/install.html) however.

Fixes #478 🦕


setup.py
```diff
@@ -85,7 +85,7 @@ def readme():
         "google-auth>=1.25.0,<3.0.0dev",  # Work around pip wack.
         "google-cloud-bigquery>=2.25.2,<4.0.0dev",
         "google-cloud-bigquery-storage>=2.0.0,<3.0.0dev",
-        "pyarrow>=3.0.0,<7.0dev",
+        "pyarrow>=3.0.0,<=9",
```
parthea, line 88:
```suggestion
        "pyarrow>=3.0.0,<10.0dev",
```

willsthompson:
Your prerelease build fails because this installs Shapely 2.0a1
```
nox > python -m pip install --prefer-binary --pre --upgrade alembic geoalchemy2 google-api-core google-cloud-bigquery google-cloud-bigquery-storage sqlalchemy shapely google-cloud-core grpcio
```
and that version introduces breaking changes from 1.x for google-cloud-bigquery. The only reference to shapely I could find was in a google-cloud-bigquery [extras definition](https://github.com/googleapis/python-bigquery/blob/543a8e4026d8aa5b6de377cc765e8aaa5279a58c/setup.py#L55), and that constrains shapely to "Shapely>=1.6.0, <2.0dev", which won't install the alpha release that's breaking this build. 

How should this build failure be resolved?

kaxil:
Any updates on this PR?

As it stands, a user can't use `snowflake-connector-python` with `sqlalchemy-bigquery 1.4.4` 

```
The conflict is caused by:
    sqlalchemy-bigquery 1.4.4 depends on pyarrow<7.0dev and >=3.0.0
    snowflake-connector-python[pandas] 2.8.0 depends on pyarrow<8.1.0 and >=8.0.0; extra == "pandas"
```

willsthompson:
This PR has been open so long that a newer PyArrow version has been released, and this PR should be updated to include it. I can do that when I get back from vacation, but is anyone from this organization available to review and comment on how to resolve the build failure? I’m happy to puzzle through dependency trees if someone can tell me what it is I actually need to fix. 

Requiring stale versions of popular packages is pathological to Python development because pip offers no mechanism to override conflicting dependencies, so this package is a bit of an albatross for anyone who needs to use it. This only gets worse as more of the PyArrow ecosystem drops support for old versions

kaxil:
@tswast ^^ Any help will be appreciated

tswast:
@chalmerlowe and I discussed this. Since PyArrow releases so many major versions that don't actually break our usage, I think it's best if we just omit the top constraint on pyarrow across all of the BigQuery packages.

tswast (approved):

willsthompson:
> I think it's best if we just omit the top constraint on pyarrow across all of the BigQuery packages.

tboddyspargo:
@willsthompson @tswast - FWI the accidental closure of this PR may have removed the “automerge” label.

chalmerlowe:
20221128: there is a known Kokoro prerelease issue with parsing the sqlalchemy version number.
The prerelease version number is currently x.x.0b2.
The patch value of 0b2 cannot be converted to an integer and thus the prerelease test fails.
For the moment, I am ignoring this issue.

PR #513, opened 2022-11-29
fix: use packaging version parser instead of string splitting

Mattix23:
🦕
Pair program with @tswast , @thejaredchapman , @aribray


tswast (approved):
Looks great!

Agreed that we should not allow SQLAlchemy 2.x for now, since there are still some prerelease failures.

PR #733, opened 2023-01-18
fix: remove type annotations from _struct.py

### LINKED ISSUE #545: Wrong types for STRUCT.__init__() args
I am getting Pylance type-checking warnings when trying to use `STRUCT`. My code looks like this:

```
...
Column(
    "analysis_filter",
    ARRAY(STRUCT(dimension=String, values=ARRAY(STRUCT(value=String)))),
),
...
```

Pylance underlines `dimension=String` and gives this warning:

```
Argument of type "Type[String]" cannot be assigned to parameter "dimension" of type "Mapping[str, TypeEngine]" in function "__init__"
  "Type[type]" is incompatible with "Type[Mapping[str, TypeEngine]]"
```

And similar warnings for `values=ARRAY(...)` and `value=String`.

I think the problem might be wrong types for the args in the `STRUCT.__init__()` function. That function is declared in _struct.py as follows:

```
def __init__(
    self,
    *fields: Tuple[str, sqlalchemy.types.TypeEngine],
    **kwfields: Mapping[str, sqlalchemy.types.TypeEngine],
):
```

But according to PEP 484, types for `*args` and `**kwargs` should be the type of a single argument, not the `Tuple` or `Dict` type of `args` or `kwargs` itself (see https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values).

So I think `**kwfields` should be declared like this:

```
    **kwfields: sqlalchemy.types.TypeEngine,
```

I don't know how `*fields` should be declared. I'm not sure whether it's possible to specify different types for the individual elements of `*args`. Maybe this function shouldn't use `*args` and should use two explicitly defined arguments instead?

aribray:
Removes type annotations from _struct.py

 Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-sqlalchemy/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
 Ensure the tests and linter pass
 Code coverage does not decrease (if any source code was changed)
 Appropriate docs were updated (if necessary)

Fixes #545  🦕


setup.py
```diff
@@ -46,7 +46,7 @@ def readme():
 
 
 extras = dict(
-    geography=["GeoAlchemy2", "shapely"],
+    geography=["GeoAlchemy2==0.12.5", "shapely"],
```
chalmerlowe, line 49: It is not clear why we included a specific version of `GeoAlchemy2` in this PR.
Ari is not supporting this task, so I am unable to ask about it.
I am gonna remove it and ensure that tests pass without it.

If tests pass, I recommend we merge.

chalmerlowe, line 49:
```suggestion
    geography=["GeoAlchemy2", "shapely"],    
```

chalmerlowe (approved):
LGTM

PR #736, opened 2023-01-25
fix: Avoid sqlalchemy20

chalmerlowe:
Prevents the prerelease/compliance installs from using sqlalchemy at >2.0.0 (NOTE: sqlalchemy 2.0.0 has many breaking changes and is not being supported at this time).

* Removes the nox `prerelease` install in deference to the autogenerated `prerelease_deps` install.
* prevents prerelease testing from installing sqlalchemy > 2.0.0
* adds a warning suppression environmental variable related to attempts by the code to use syntax that is not found in sqlalchemy 2.0.0
* corrects an error in how dependency testing is handled (by eliminating the `SYSTEM_TEST_EXTRAS` collection of test modules (specifically the only package in this collection was being improperly installed)).


shollyman (commented):
Rather than removing these configs, should we instead disable the kokoro prerelease target internally?

noxfile.py
```diff
@@ -369,7 +299,7 @@ def compliance(session):
         session.skip("Compliance tests were not found")
 
     session.install("--pre", "grpcio")
-
+    session.install("--pre", "--no-deps", "--upgrade", "sqlalchemy<2.0.0")
```
shollyman, line 302: the goal with the upgrade + constraint is to pick up the latest of everything but sqlalchemy?

owlbot.py
```diff
@@ -72,10 +72,28 @@
...
 
+
+s.replace(
+    ["noxfile.py"], 
+    "\+ SYSTEM_TEST_EXTRAS",
+    "",
+)
+
+
+s.replace(
```
shollyman, line 88: I always worry about brittleness with these kinds of replacements, but alas that's problem for future us.

shollyman (approved):
LGTM, but I defer to parthea who's far more familiar with the operational aspects here.

noxfile.py
chalmerlowe: The constraint is to ensure that we install a prerelease version of sqlalchemy up to but not including 2.0.0.

owlbot.py
chalmerlowe: you and me both. 
i really disliked these `s.replace()` functions as I was writing them.

parthea (approved):

PR #737, opened 2023-01-26
docs: pass credentials as python dictionary

### LINKED ISSUE #331: Can we pass the credentials directly as dictionary/json instead of key json file path?
Actually I want to take credentials as input from the user and saving it in the file would be not appropriate, so is there a way where i can directly pass credentals to connect method instead of json key file path?

alvarowolfx:
Document how to pass the credentials directly as dictionary/json instead of key json file path.

Fixes #331  🦕


README.rst
```diff
@@ -104,12 +104,21 @@ Project
...
 
 .. code-block:: python
 
+    # provide the path to a service account JSON file
     engine = create_engine('bigquery://', credentials_path='/path/to/keyfile.json')
 
+    # provide credentials as a Python dictionary
+    credentials_info = {
+        "type": "service_account", 
+        "project_id": "service-account-project"
```
chalmerlowe, line 119: In examples, my sense is that it sometimes helps to highlight for the end user (especially a novice) that they need to replace something by using a phrase that indicates they should provide the value (i.e. using the term "your").

In the following example I don't know if `service_account` is intended to be literal text or if we intend for the user to supply a value. If it is supposed to be user supplied, I think that should be clarified.

I am presuming that the `service-account-project` is intended to be user-specific, so I recommend a change such as the following:

```suggestion
        "type": "service_account", 
        "project_id": "your-service-account-project-identifier"
```

m-strzelczyk (approved):

alvarowolfx:
@chalmerlowe thanks for the review. I agree your recommendations and updated the PR accordingly.

chalmerlowe (approved):
LGTM

Issue #368, opened 2021-10-28
Unnest Function In BigQuery SQLAlchemy Plugin Joins Twice When Selecting Other Columns in CTE or Subquery 


mgtcardenas:
From a customer support request:

**Issue summary**: This issue is closely related to [Unnest function in BigQuery SQLAlchemy plugin joins twice when selecting other columns #353](https://github.com/googleapis/python-bigquery-sqlalchemy/issues/353). After the previous issue was resolved, queries using the UNNEST function that are not in Common Table Expressions (CTE) or subqueries work as expected. See the following minimal example:

```
from sqlalchemy import create_engine, MetaData, select, func as F, Table, Column, Integer, ARRAY

metadata = MetaData()
table = Table(
  "table1",
  metadata,
  Column("foo", Integer),
  Column("bars", ARRAY(Integer))
)

engine = create_engine("bigquery://", future=True)
q = select(table.c.foo, F.unnest(table.c.bars).column_valued("bar"))
print(q.compile(engine))
```

This outputs the following, which is expected:
```
SELECT `table1`.`foo`, `bar` 
FROM `table1`, unnest(`table1`.`bars`) AS `bar`
```

However, if we use the table as a subquery or CTE, the issue manifests again. The following code (continuing from above):

```
q = q.cte("cte")
q = select(*q.columns)
print(q.compile(engine))
```

Produces the next output:

```
WITH `cte` AS 
(SELECT `table1`.`foo` AS `foo`, `bar` 
FROM `table1`, `table1` `table1_1`, unnest(`table1_1`.`bars`) AS `bar`)
 SELECT `cte`.`foo`, `cte`.`bar` 
FROM `cte`
```
Note that `table1` is duplicated in the `FROM` clause again.

**Expected behavior**: It is expected to generate the following query instead (note the different FROM clause):
```
WITH `cte` AS 
(SELECT `table1`.`foo`, `bar` 
FROM `table1`, unnest(`table1`.`bars`) AS `bar`)
 SELECT `cte`.`foo`, `cte`.`bar` 
FROM `cte`
```

tswast:
@kwohlfahrt thanks for the examples and the note regarding Postgres.

lazjake:
👋 any progress on this one? sadly running into the nested cte issue 😢

set92:
Since this look completely stagnant, someone knows some workaround it? Is possible to write the `UNNEST(table.col) AS x` as text in the FROM?

[...]

PR #839, opened 2023-02-22
fix: Avoid aliasing known CTEs

snapiri:
I now also add the Common Table Expression (CTE) names to the known tables, so they will not be re-added (thus cross-joined with self).

The following snippet (from the issue) was used to validate the change:
```
    from sqlalchemy import create_engine, MetaData, select, func as F, Table, Column, Integer, ARRAY

    metadata = MetaData()

    table = Table(
        "table1",
        metadata,
        Column("foo", Integer),
        Column("bars", ARRAY(Integer))
    )

    selectable = sql.select(table.c).select_from(table).cte('cte')
    q = sql.select([selectable.c.foo, sql.func.unnest(selectable.c.bars).column_valued('unnest_bars')]).select_from(selectable)
    print(q.compile(engine))
```

---

- [X] Appropriate docs were updated (if necessary)

Related #368


tswast:
Kokoro failure looks like https://github.com/googleapis/python-bigquery-sqlalchemy/issues/875

We might need to address that before we can merge.

sqlalchemy_bigquery/base.py
```diff
@@ -261,6 +261,7 @@ def _known_tables(self):
             if isinstance(from_, Table):
                 known_tables.add(from_.name)
             elif isinstance(from_, CTE):
+                known_tables.add(from_.name)
```
tswast, line 264: Any possibility you could add a test that was failing before this change? Or at least provide a code example so that one of the package maintainers could add such a test.

snapiri: Unit tests added

tswast (approved):

PR #847, opened 2023-03-17
feat: Remove pyarrow and bqstorage as dependencies

### LINKED ISSUE #743: pyarrow dependency should be optional
pyarrow is an [optional dependency](https://github.com/googleapis/python-bigquery/blob/204779c57e706c002ababff72f8e45f1e376b960/setup.py#L31-L69) of google-cloud-bigquery, but it's [made mandatory](https://github.com/googleapis/python-bigquery-sqlalchemy/blob/a3b8a5f0bbea79b11d5390aef547cfd87bb5808c/setup.py#L90) by python-bigquery-sqlalchemy.

As pyarrow is quite large on disk &mdash; 100 MB on x86_64 Linux &mdash; I don't want to install it when it's unused in my application (AFAICT I don't need `google-cloud-bigquery-storage` either but that's [not huge](https://pypi.org/project/google-cloud-bigquery-storage/#files)).

I suggest:

1. Removing the direct dependencies on `pyarrow` and `google-cloud-bigquery-storage`
2. Add a `bqstorage` extra that depends on `google-cloud-bigquery[bqstorage]`. That'll respect upstream's version bounds without introducing local bounds that could cause conflicts for users.
3. Document that users wanting improved performance with large result sets should install `bigquery-sqlalchemy[bqstorage]`.

There's an existing PR at #470 but it looks like it has stalled out, so I'm filing this issue to provide a blueprint for someone who's able to do this work.

chalmerlowe:
Removes `pyarrow` and `bqstorage` as regular dependencies and moves them to the extras category where they can be installed, as needed (i.e. when the use of `pyarrow` and/or `bqstorage` may provide performance benefits).

As noted in Issue #743:

`pyarrow` is an [optional dependency](https://github.com/googleapis/python-bigquery/blob/204779c57e706c002ababff72f8e45f1e376b960/setup.py#L31-L69) of `google-cloud-bigquery`, but it's [made mandatory](https://github.com/googleapis/python-bigquery-sqlalchemy/blob/a3b8a5f0bbea79b11d5390aef547cfd87bb5808c/setup.py#L90) by `python-bigquery-sqlalchemy`.

Because `pyarrow` is quite large on disk — 100 MB on x86_64 Linux — installing it when it is not used is inefficient and poses a burden on some installations in space constrained environments.
 
This is not deemed a breaking change in that no existing functionality is being deprecated and existing users still have the ability to install `bqstorage`/`pyarrow` as needed. 

NOTE: Users are likely to benefit from installing `bqstorage` when accessing query results in excess of 100 Mb ... these users may experience a speed increase up to 10x. 

Fixes #743 🦕

IF this PR is merged, we should be able to close the following adjacent/associated PR:

Closes https://github.com/googleapis/python-bigquery-sqlalchemy/pull/470


tswast (commented):
Thanks! This approach LGTM.

noxfile.py
```diff
@@ -209,6 +221,13 @@ def unit(session):
     default(session)
 
 
+@nox.session(python=[UNIT_TEST_PYTHON_VERSIONS[0], UNIT_TEST_PYTHON_VERSIONS[-1]])
+def unit_noextras(session):
```
tswast, line 225: Given that we run unit test on more than just 3.8 and 3.11, I think we can drop this session.

chalmerlowe: @tswast 
The unit tests run with the extras installed.
This session is aimed at running the same tests when the extras are NOT installed.
On the unit testing level, do we want an `extras` installed vs `no extras` installed comparison?

tswast: UNIT_TEST_EXTRAS_BY_PYTHON only has entries for Python 3.8 and 3.11. 

UNIT_TEST_PYTHON_VERSIONS has 3.7 through 3.11.

That means we run with no extras (besides those needed for tests) on 3.7, 3.9, and 3.10.

I figure that should be sufficient, especially given that UNIT_TEST_PYTHON_VERSIONS[0] is 3.7

tswast (commented):
LGTM once you get owlbot sorted.

chalmerlowe:
@tswast I got owlbot to work and all the tests are passing.
If you approve, we can merge this PR.

tswast (approved):
Thanks!

PR #900, opened 2023-08-24
fix: remove Python 3.7 support due to EOL

nayaknishant:
Removing Python 3.7 support from repo due to EOL (https://devguide.python.org/versions/)


dandhlee:
Can you also send a CL to remove the job configs?

nayaknishant:
Python 3.7 removed as a required check in https://github.com/googleapis/python-bigquery-sqlalchemy/pull/902. Samples - Python 3.7 will be failing in this PR but will be removed as a check in future PRs.

dandhlee (approved):

tswast:
Please make sure you update the `testing/constraints-*.txt` files so that the minimum version of Python we test with is testing with the minimum versions of our dependencies that we claim to support in `setup.py`.

PR #1009, opened 2024-01-18
test: prevent alchemy test suite from being rate-limited.

nlenepveu:
Checking the error message in the Kokoro Prerelease Dependencies check in https://github.com/googleapis/python-bigquery-sqlalchemy/pull/936, it appears that the alembic system test suite is beeing rate-limited after exceeding the DDL operations quota (https://cloud.google.com/bigquery/quotas#standard_tables).

I suggest we could add a delay to prevent the test from being rate-limited. This is not very elegant but it should not have too much impacts on the overall delay of the test suite.

We could also introduce a more relevant retry-strategy that would requires some more work.

----

### DIFF
--- a/tests/system/test_alembic.py
+++ b/tests/system/test_alembic.py
@@ -107,6 +107,18 @@ def test_alembic_scenario(alembic_table):
         SchemaField("last_transaction_date", "DATETIME", description="when updated"),
     ]
 
+    op.rename_table("account", "accounts")
+    assert alembic_table("account") is None
+    assert alembic_table("accounts", "schema") == [
+        SchemaField("id", "INTEGER", "REQUIRED"),
+        SchemaField("name", "STRING(50)", "REQUIRED", description="The name"),
+        SchemaField("description", "STRING(200)"),
+        SchemaField("last_transaction_date", "DATETIME", description="when updated"),
+    ]
+
+    op.drop_table("accounts")
+    assert alembic_table("accounts") is None
+
     op.create_table(
         "account_w_comment",
         Column("id", Integer, nullable=False),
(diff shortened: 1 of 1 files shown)


tests/system/test_alembic.py
```diff
@@ -146,23 +158,39 @@ def test_alembic_scenario(alembic_table):
...
 
+    _prevent_rate_limiting()
     op.drop_table("identifiers")
+
+
+def _prevent_rate_limiting():
+    """
+    Adds a delay between operations to prevent exceeding the "Maximum rate of table metadata update operations per table".
+
+    The rate limit is set to a maximum of 5 operations per 10 seconds. By introducing a sleep of 2 seconds
```
tswast, line 193: I'm somewhat surprised that our exponential backoff query job retry logic doesn't handle this already. Do we have a link to a failing test stacktrace?

Instead of sleep, I'd like for us to add the rate limit as an allowed job retry reason in

https://github.com/googleapis/python-bigquery/blob/d1161dddde41a7d35b30033ccbf6984a5de640bd/google/cloud/bigquery/retry.py#L76-L84

if we know that the error is safe to retry.

If we do go this route, let's add a link to https://cloud.google.com/bigquery/quotas#standard_tables where this limit is defined.

nlenepveu: @tswast 

Sure, it would be best handled in the `python-bigquery` code.. The work around I suggested is a quite quick and dirty solution..

Here's a stacktrace of the error https://source.cloud.google.com/results/invocations/ffafb866-6bc0-423f-a86b-df69fb270d57/log.

tswast: I've filed https://github.com/googleapis/python-bigquery/issues/1790 as the preferred approach to fixing this problem.

chalmerlowe: I am inclined to see if we can get this fixed.
Will work through things on the Issue Tim created...https://github.com/googleapis/python-bigquery/issues/1790

nlenepveu:
This is now obsolete, as @kiraksi has resolved the underlying retry strategy issue in `python-bigquery`. (Issue: https://github.com/googleapis/python-bigquery/issues/1790 / PR: https://github.com/googleapis/python-bigquery/pull/1794)

PR #1048, opened 2024-03-24
fix: implement modulus operator

aholyoke:

---
Adds implementation of mod operator `%`

Currently if I try to use `%` in python code, it will fall back to the SA default compiler implementation which renders it as a percent sign, which is not syntactically valid in BQ. I could workaround this by using `func.mod`, but this isn't portable across dialects (eg. SQL Server has support for `%`, but not `MOD`)


aholyoke:
@Linchin This seems like a pretty safe change. Any chance you can review it?

chalmerlowe:
@aholyoke 

Can you do me a favor?
Can you provide me with a example of what the outcome would look like if used *in situ* (not the fake version generated by the test code) but a sample that includes the calculation of a modulo compared to zero for example:

The sqlalchemy commands that would generate a statement that includes something akin to `blah, blah, blah MOD(X, Y) == 0`

I want to ensure that I am picturing this the way you are.

Thanks.

aholyoke:
@chalmerlowe Sure, this is how it behaves before the change
```python
In [1]: import sqlalchemy as sa; from sqlalchemy_bigquery import BigQueryDialect

In [2]: meta = sa.MetaData(); table = sa.Table("table_name", meta, sa.Column("col_name", sa.Integer()))

In [3]: print(str(sa.select(table.c.col_name).where(table.c.col_name % 2 == 0).compile(dialect=BigQueryDialect())))
SELECT `table_name`.`col_name`
FROM `table_name`
WHERE `table_name`.`col_name` % :col_name_1 = :param_1
```
Which the following screenshots confirm is not the correct syntax in BQ:
Screenshot 2024-06-30 at 9 58 36 PM
Screenshot 2024-06-30 at 9 58 53 PM


and this is how it behaves after the change
```python
In [1]: import sqlalchemy as sa; from sqlalchemy_bigquery import BigQueryDialect

In [2]: meta = sa.MetaData(); table = sa.Table("table_name", meta, sa.Column("col_name", sa.Integer()))

In [3]: print(str(sa.select(table.c.col_name).where(table.c.col_name % 2 == 0).compile(dialect=BigQueryDialect())))
SELECT `table_name`.`col_name`
FROM `table_name`
WHERE MOD(`table_name`.`col_name`, :col_name_1) = :param_1

In [4]: sa.select(table.c.col_name).where(table.c.col_name % 2 == 0).compile(dialect=BigQueryDialect()).params
Out[4]: {'col_name_1': 2, 'param_1': 0}
```

chalmerlowe (approved):
LGTM.

chalmerlowe:
@aholyoke 

Thanks for pushing this PR. We appreciate the effort and your support in better understanding the nature of the issue this solves. 

Have a great day! 💜

PR #1100, opened 2024-07-31
feat: adds user agent parameters to two functions

### LINKED ISSUE #1082: Add ability to set custom User Agent
The Google Cloud Ready - BigQuery partnership program requires partners to set a custom UserAgent. This is so Google can correctly attribute usage. However, there is no built in way to set the UserAgent in this library. It's hardcoded to `sqlalchemy` right now.

https://github.com/googleapis/python-bigquery-sqlalchemy/blob/0b3ab54c368fecd148eaa101c7939e10f3cf3b05/sqlalchemy_bigquery/_helpers.py#L27-L69

The `create_bigquery_engine` method should also accept a user_agent and use that instead of the default sqlalchemy user_agent, if provided.

For now, as a workaround in my code, I'm reimplementing the `create_bigquery_engine` method myself and setting the client_info directly.

chalmerlowe:
This adds:

* the ability for users to set a customer user agent (instead of the default user agent string that is currently hard coded in)
* docstrings and type hinting for the functions that we touched
* tests to test the new code path in `google_client_info`.

- [x] Add tests to check various states/edge cases (TBD)

Fixes #1082  🦕

### DIFF
--- a/sqlalchemy_bigquery/_helpers.py
+++ b/sqlalchemy_bigquery/_helpers.py
@@ -45,7 +45,7 @@ def create_bigquery_client(
     default_query_job_config: Optional[google.cloud.bigquery.job.QueryJobConfig] = None,
     location: Optional[str] = None,
     project_id: Optional[str] = None,
-    user_agent: Optional[str] = None,
+    user_agent: Optional[google.api_core.client_info.ClientInfo] = None,
 ) -> google.cloud.bigquery.Client:
     """Construct a BigQuery client object.
 
--- a/sqlalchemy_bigquery/_helpers.py
+++ b/sqlalchemy_bigquery/_helpers.py
@@ -6,6 +6,7 @@
 
 import functools
 import re
+from typing import Optional
 
 from google.api_core import client_info
 import google.auth
@@ -14,6 +15,7 @@ from google.oauth2 import service_account
 import sqlalchemy
 import base64
 import json
+
 
 
 USER_AGENT_TEMPLATE = "sqlalchemy/{}"
--- a/sqlalchemy_bigquery/_helpers.py
+++ b/sqlalchemy_bigquery/_helpers.py
@@ -24,19 +24,47 @@ SCOPES = (
 )
 
 
-def google_client_info():
-    user_agent = USER_AGENT_TEMPLATE.format(sqlalchemy.__version__)
+def google_client_info(user_agent=None):
+    """
+    Return a client_info object, with an optional user agent
+    string.  If user_agent is None, use a default value.
+    """
+
+    if user_agent is None:
+        user_agent = USER_AGENT_TEMPLATE.format(sqlalchemy.__version__)
     return client_info.ClientInfo(user_agent=user_agent)
 
 
 def create_bigquery_client(
-    credentials_info=None,
-    credentials_path=None,
-    credentials_base64=None,
-    default_query_job_config=None,
-    location=None,
-    project_id=None,
-):
+    credentials_info: Optional[dict] = None,
+    credentials_path: Optional[str] = None,
+    credentials_base64: Optional[str] = None,
+    default_query_job_config: Optional[google.cloud.bigquery.job.QueryJobConfig] = None,
+    location: Optional[str] = None,
+    project_id: Optional[str] = None,
+    user_agent: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


sqlalchemy_bigquery/_helpers.py
```diff
@@ -24,19 +25,46 @@
...
+        credentials_path Optional[str]:
+        credentials_base64 Optional[str]:
+        default_query_job_config (Optional[google.cloud.bigquery.job.QueryJobConfig]):
+            Default ``QueryJobConfig``.
+            Will be merged into job configs passed into the ``query`` method.
+        location (Optional[str]):
+            Default location for jobs / datasets / tables.
+        project_id (Optional[str]):
+            Project ID for the project which the client acts on behalf of.
+        user_agent (Optional[google.api_core.client_info.ClientInfo]):
```
Linchin, line 61: NIT:

Looks like it should be string?

```suggestion
        user_agent (Optional[str]):
```

sqlalchemy_bigquery/_helpers.py
```diff
@@ -24,19 +25,46 @@
 )
 
 
-def google_client_info():
-    user_agent = USER_AGENT_TEMPLATE.format(sqlalchemy.__version__)
+def google_client_info(user_agent=None):
```
Linchin, line 28: NIT:

Consider this isn's a private method, maybe we can add type hint for this parameter too?

Linchin:
LGTM. Just some thoughts: should we add a system test to verify that user agent is indeed successfully set up?

Linchin (approved):

sqlalchemy_bigquery/_helpers.py
chalmerlowe: This line is correct. The other line that had `str` was in error.

`google_client_info` returns a `ClientInfo` object which may include the `user_agent` string as well as several other attributes:

```
self.python_version = python_version
self.grpc_version = grpc_version
self.api_core_version = api_core_version
self.gapic_version = gapic_version
self.client_library_version = client_library_version
self.user_agent = user_agent
self.rest_version = rest_version
```        

In addition, looking at the `bigquery.Client` object, it is expecting a ClientInfo object:

```
client_info (Optional[google.api_core.client_info.ClientInfo]):
```

sqlalchemy_bigquery/_helpers.py
chalmerlowe: Done.

sqlalchemy_bigquery/_helpers.py
Linchin: Just looking at the code, `user_agent` is passed to `google_client_info()`, which is expecting a string. `google.api_core.client_info.ClientInfo` seems to be what `google_client_info()` returns instead?

chalmerlowe:
> LGTM. Just some thoughts: should we add a system test to verify that user agent is indeed successfully set up?

I am not inclined to put the time/resources into this minor change.

Right now, when the `to_user_agent()` method is called...the customer-supplied user agent string has a variety of elements appended to it. These elements are dynamic and change depending on the versions of certain components that are installed.

Are you picturing an end-to-end system test that accounts for each element of this string? If so, then that type of test feels like it may be brittle and would be dependent on exactly what sub-components are present in this function and might involve complex checks like regex. There might be ways to simplify the checks. 

If you have alternate ideas on how to do this simply, easily, and in a way that is robust to change within the code, you are welcome to suggest something.

```
if self.grpc_version is not None:
    ua += "grpc/{grpc_version} "

if self.rest_version is not None:
    ua += "rest/{rest_version} "

    ua += "gax/{api_core_version} "

if self.gapic_version is not None:
    ua += "gapic/{gapic_version} "

if self.client_library_version is not None:
    ua += "gccl/{client_library_version} "

```

Linchin:
I will add a system test similar to the rest of tests in [test_helpers.py](https://github.com/googleapis/python-bigquery-sqlalchemy/blob/80781ef99287af2e950f21ca399c84d20422b732/tests/system/test_helpers.py). Sorry that I'm unable to suggest code change to locations unedited by this PR, apparently it's a [feature request with github](https://github.com/orgs/community/discussions/4452) that hasn't been fulfilled. Feel free to modify as needed.

PR #1113, opened 2024-09-11
Feat: Update partitioning by DATE, DATETIME, TIMESTAMP, _PARTITIONDATE

chalmerlowe:
This adds some additional functionality to properly handle partitioning of columns with the following datatypes.  

* DATE
* TIMESTAMP
* DATETIME
* _PARTITIONDATE

Where appropriate, ensures the following functions can be used with/or without the following TimePartitioningTypes (HOUR, DAY, MONTH, YEAR).

* DATE_TRUNC()
* TIMESTAMP_TRUNC()
* DATETIME_TRUNC()
* DATE()

This is a nearly complete fix for googleapis/google-cloud-python#15782. A revised table from googleapis/google-cloud-python#15782 is included here:
NOTES: 

* This PR does not handle _PARTITIONTIME
* If a field (column) is not provided to partition on, the system defaults to the pseudocolumn "_PARTITIONDATE" to partition by. (If you attempt to provide a partitioning period anyway BQ does not accept HOUR and even if you include DAY, MONTH, or YEAR, the system will ignore those in deference to the _PARTITIONDATE).

Column Data Type | HOUR | DAY | MONTH | YEAR | Special case: None |
--------------- | -- | -- | -- | -- | -- |
DATE | N/A | N/A | via DATE_TRUNC | via DATE_TRUNC | N/A | 
DATETIME | via DATETIME_TRUNC | via DATETIME_TRUNC | via DATETIME_TRUNC | via DATETIME_TRUNC | via DATETIME_TRUNC (defaults to DAY) |
TIMESTAMP | via TIMESTAMP_TRUNC | via TIMESTAMP_TRUNC | via TIMESTAMP_TRUNC | via TIMESTAMP_TRUNC |  N/A |
_PARTITIONDATE | N/A | N/A | N/A | N/A | N/A |  
_PARTITIONTIME | Not currently implemented TODO: USE TIMESTAMP_TRUNC | Not currently implemented TODO: USE TIMESTAMP_TRUNC | Not currently implemented TODO: USE TIMESTAMP_TRUNC | Not currently implemented TODO: USE TIMESTAMP_TRUNC | Need to lookup |


sqlalchemy_bigquery/base.py
```diff
@@ -830,23 +832,61 @@ def _process_time_partitioning(
...
         """
+
         field = "_PARTITIONDATE"
         trunc_fn = "DATE_TRUNC"
 
+        # Format used with _PARTITIONDATE which can only be used for
+        # DAY / MONTH / YEAR
+        if time_partitioning.field is None and field == "_PARTITIONDATE":
+            if time_partitioning.type_ in {"DAY", "MONTH", "YEAR"}:
+                return f"PARTITION BY {trunc_fn}({field})"
```
suzmue, line 857: I'm a little confused, the type isn't passed to the trunc_fn, should it be?

sqlalchemy_bigquery/base.py
```diff
@@ -830,23 +832,61 @@ def _process_time_partitioning(
...
+        * DATE(_PARTITIONTIME)
+        * DATE(<timestamp_column>)
+        * DATE(<datetime_column>)
+        * DATE column
         """
+
         field = "_PARTITIONDATE"
         trunc_fn = "DATE_TRUNC"
 
+        # Format used with _PARTITIONDATE which can only be used for
```
suzmue, line 853: There are a lot of cases here that make this function hard to parse. In particular, I think it could be improved if instead of the first special casing for time.partitioning.field is None block was instead combined by deleting that check, and moving the if time.partioning.type_ in ... below the if field is not None block:

```python
        field = "_PARTITIONDATE"
        trunc_fn = "DATE_TRUNC"
        if time_partitioning.field is not None:
            field = time_partitioning.field

            if isinstance(
                table.columns[field].type,
                (sqlalchemy.sql.sqltypes.TIMESTAMP),
            ):
                trunc_fn = "TIMESTAMP_TRUNC"
            elif isinstance(
                table.columns[field].type,
                sqlalchemy.sql.sqltypes.DATETIME,
            ):
                  trunc_fn = "DATETIME_TRUNC"



        if trunc_fn == "DATE_TRUNC" and not time_partitioning.type_ in {"DAY", "MONTH", "YEAR"}:
               raise ValueError(
                      f"DATE_TRUNC can only be used with TimePartitioningTypes {{DAY, MONTH, YEAR}} received {time_partitioning.type_}"
                  )

        # Format used with generically with DATE, TIMESTAMP, DATETIME, DATE_TRUNC
        return f"PARTITION BY {trunc_fn}({field}, {time_partitioning.type_})"**
```

suzmue: I had removed the the last `if isinstance(table.columns[field].type,`, because I thought it was part of the if elif, I see now that its separate. If thats the case, it could be the first if condition so it can take priority instead of setting it back after the fact (pattern matching would be nice here when we support >= python 3.10)

sqlalchemy_bigquery/base.py
chalmerlowe: Been looking at your comments related to partitioning.

My original code doesn't seem to be 100%, so glad you asked some questions. = )
I am gonna revisit the code and will try to make it correct and simpler, if possible.

Our goal is to increase our coverage of use cases to four.

Within each of those four, the SQL signature will look different depending on the use case and the associated function AND the allowable `TimePartitioningType`.

Some things to note: _PARTITIONDATE does not allow a function OR a `TimePartitioningType` at all.
Some functions only take a couple of `TimePartitioningTypes` (TPT). See the breakdown below.

```
# _PARTITIONDATE has no function, no TPT

CREATE TABLE `experimental.some_table` ( `id` INT64, `createdAt` DATE )
PARTITION BY _PARTITIONDATE;

# DATETIME_TRUNC has function and four TPTs

CREATE TABLE `experimental.some_table` ( `id` INT64, `createdAt` DATETIME )
DATETIME_TRUNC(<datetime_column>, DAY/HOUR/MONTH/YEAR); 
 
# TIMESTAMP_TRUNC has function and four TPTs

CREATE TABLE `experimental.some_table` ( `id` INT64, `createdAt` TIMESTAMP )
TIMESTAMP_TRUNC(<timestamp_column>, DAY/HOUR/MONTH/YEAR);

# DATE_TRUNC has function and only two TPTs

CREATE TABLE `experimental.some_table` ( `id` INT64, `createdAt` DATE )
PARTITION BY DATE_TRUNC(createdAt, MONTH/YEAR);
```

sqlalchemy_bigquery/base.py
chalmerlowe: @suzmue 

I am considering an approach such as this. It separates each decision point into concrete steps and incrementally leads to a final choice.

It should be easily expandable later when/if we add additional conditions (i.e. _PARTITIONTIME or the use of the DATE() function with specific column types).

Using a mapping such as `sqltypes` also puts all the elements of the truth table in one concise spot (i.e. if using TIMESTAMP, the function will be TIMESTAMP_TRUNC and the allowable partition_periods are D/H/M/Y) which is the important part and allows us to have very simple test conditions and dictionary lookup in the code.

NOTE: I have not tested this. Consider this pseudocode while we talk through the concept.
 
```
from sqlalchemy.sql.sqltypes import (
    TIMESTAMP,
    DATETIME,
    DATE,
)

sqltypes = {
    _PARTITIONDATE: ("_PARTITIONDATE", None),
    TIMESTAMP: ("TIMESTAMP_TRUNC", {DAY, HOUR, MONTH, YEAR}),
    DATETIME: ("DATETIME_TRUNC", {DAY, HOUR, MONTH, YEAR}),
    DATE: ("DATE_TRUNC", {DAY, MONTH, YEAR}),
}

# Extract field (i.e <column_name> or _PARTITIONDATE)
# AND extract the column_type (i.e. TIMESTAMP, DATE, DATETIME, _PARTITIONDATE)
if time_partitioning.field is not None:
    field = time_partitioning.field
    column_type = table.columns[field].type

else:
    field = "_PARTITIONDATE"
    column_type = _PARTITIONDATE

# Extract time_partitioning.type_ (DAY, HOUR, MONTH, YEAR)
# i.e. generates one partition per type (1/DAY, 1/HOUR)
if time_partitioning.type_ is not None:
    partitioning_period = time_partitioning.type_

# Extract the truncation_function (i.e. DATE_TRUNC)
# and the set of allowable partition_periods 
# that can be used in that function
trunc_fn, allowed_partitions = sqltypes[column_type]

# Create output:

# Special case: _PARTITIONDATE does NOT use a function or partitioning_period
if trunc_fn == "_PARTITIONDATE":
    return f"PARTITION BY {field}"

# Generic case 
if partitioning_period not in allowed_partitions:
    raise ValueError("error msg")

return f"PARTITION BY {trunc_fn}({field}), {partitioning_period}"
```

suzmue: Your approach lgtm!

tests/unit/test_table_options.py
```diff
@@ -132,83 +188,89 @@ def test_table_require_partition_filter_dialect_option(faux_conn):
             bigquery_require_partition_filter=True,
         )
 
-    assert " ".join(faux_conn.test_data["execute"][-1][0].strip().split()) == (
+    result = " ".join(faux_conn.test_data["execute"][-1][0].strip().split())
```
Linchin, line 191: NIT:

It seems this test is covered in `test_table_time_partitioning_date_timestamp_and_datetime_dialect_option`?

sqlalchemy_bigquery/base.py
```diff
@@ -812,42 +812,99 @@ def _raise_for_type(self, option, value, expected_type):
...
+        if trunc_fn == "_PARTITIONDATE":
+            return f"PARTITION BY {field}"
+
+        # Special Case: BigQuery will not accept DAY as partitioning_period for
+        # DATE_TRUNC.
+        # However, the default argument in python-bigquery for TimePartioning
+        # is DAY. This case overwrites that to avoid making a breaking change in
+        # python-bigquery.
+        # https://github.com/googleapis/python-bigquery/blob/a4d9534a900f13ae7355904cda05097d781f27e3/google/cloud/bigquery/table.py#L2916
+        if trunc_fn == "DATE_TRUNC" and partitioning_period == "DAY":
```
Linchin, line 891: I guess this would be added when you complete the draft, I think we need to set the default value of `partitioning_period` somewhere

chalmerlowe: @Linchin 

I am unclear on why we might need to create a default value of `partitioning_period`. It is set based on the value of `time_partitioning.type_` on `line 873`.

`time_partitioning.type_` is set in the `python-bigquery` library in the [TimePartitioning class](https://github.com/googleapis/python-bigquery/blob/fffe6ba6b271180b0c59c9fbf70feb7d6d322906/google/cloud/bigquery/table.py#L2949C2-L2956C31).

When that class is instantiated, the default for `type_` in the `__init__` is `None` but it is then processed and immediately overwritten by `DAY` as shown at the link above.

So by the time `.type_` gets to us, it will either be `DAY` or will be set to something else that we use to set the variable named `partitioning_period`.

NOTE: I rename the variable to `partitioning_period` simply to make this code more readable. `type_` seemed less intuitive than `partitioning_period`.

tests/unit/test_table_options.py
chalmerlowe: Deduplicated.

chalmerlowe:
@Linchin Please take a look. I have taken this out of draft. As far as I recall, this should be finished and ready for final review/approval.

sqlalchemy_bigquery/base.py
Linchin: Thank you, I missed the part in `__init__()` that sets the default value.

tests/unit/test_table_options.py
```diff
@@ -104,111 +104,151 @@ def test_table_clustering_fields_dialect_option_type_error(faux_conn):
...
             sqlalchemy.Column("id", sqlalchemy.Integer),
-            sqlalchemy.Column("createdAt", sqlalchemy.TIMESTAMP),
+            sqlalchemy.Column("createdAt", sqlalchemy.DateTime),
             bigquery_time_partitioning=TimePartitioning(field="createdAt"),
         )
-
-    assert " ".join(faux_conn.test_data["execute"][-1][0].strip().split()) == (
-        "CREATE TABLE `some_table` ( `id` INT64, `createdAt` TIMESTAMP )"
-        " PARTITION BY TIMESTAMP_TRUNC(createdAt, DAY)"
+    result = " ".join(faux_conn.test_data["execute"][-1][0].strip().split())
```
Linchin, line 195: It seems the second half of this test does the following, as the docstring suggests:
```
Confirms that if the column datatype is DATETIME but no TimePartitioning.type_
    has been supplied, the system will default to DAY.
```
Maybe I'm missing something, but isn't this covered in the test case in line 136?

tests/unit/test_table_options.py
```diff
@@ -104,111 +104,151 @@ def test_table_clustering_fields_dialect_option_type_error(faux_conn):
...
-            bigquery_time_partitioning=TimePartitioning(field="createdAt"),
+            sqlalchemy.Column("createdAt", column_dtype),
+            bigquery_time_partitioning=TimePartitioning(type_=time_partitioning_type),
         )
 
     # confirm that the following code creates the correct SQL string
-    assert " ".join(faux_conn.test_data["execute"][-1][0].strip().split()) == (
-        "CREATE TABLE `some_table_2` ( `id` INT64, `createdAt` DATE )"
-        " PARTITION BY createdAt"
+    result = " ".join(faux_conn.test_data["execute"][-1][0].strip().split())
```
Linchin, line 244: Same as https://github.com/googleapis/python-bigquery-sqlalchemy/pull/1113/files#r1852967358

tests/unit/test_table_options.py
chalmerlowe: The difference between this test and the parametrized test above is minor.

Essentially, in the parameters on line 136 we explicitly provide a value of `None` as a `type_` argument to the TimePartitioning class.

We know the default is `None` in `python-bigquery` but we set the value deliberately.

In line 193, we do not set the value of `type_` explicitly, we allow the default value (which should be `None`) come through and thus confirm that everything works as expected whether we provide a value to `type_` or not.

tests/unit/test_table_options.py
chalmerlowe: In this test, we do NOT provide a `field` argument to the TimePartitioning class whereas in the tests above, we do.

This is needed to confirm that in the absence of a `field` argument the SQL statement will not use a function, etc but will instead use `_PARTITIONDATE`

Both the tests above end up including a truncation function but this on does not (and should not).

Linchin (approved):
LGTM, thank you!

Two small suggestions, could we:
1. rename the tests to reflect what they are intended to test
2. the table in the first comment seems a little outdated, could we update it?

PR #1116, opened 2024-09-19
fix: Support partitioning by DATE or Date

### LINKED ISSUE #1115: Time partitioning on Date column fails with type_="DAY"
Not exactly a duplicate of #1056! The fix for that issue copied the standard that was present of checking vs DATE, but `sa.Date` should also be valid and weirdly enough changing that is enough to break things, with the same error as #1056 regarding `DATE_TRUNC` (since the `if` block that was added does not catch `sa.Date`, just `sa.DATE`).

cc @chalmerlowe this seems potentially relevant to your work in #1113

### LINKED ISSUE #15776: Time partitioning on Date column fails with type_="DAY"
Not exactly a duplicate of googleapis/python-bigquery-sqlalchemy#1056! The fix for that issue copied the standard that was present of checking vs DATE, but `sa.Date` should also be valid and weirdly enough changing that is enough to break things, with the same error as googleapis/python-bigquery-sqlalchemy#1056 regarding `DATE_TRUNC` (since the `if` block that was added does not catch `sa.Date`, just `sa.DATE`).

cc @chalmerlowe this seems potentially relevant to your work in googleapis/python-bigquery-sqlalchemy#1113

bnaul:
Fixes googleapis/google-cloud-python#15776 and adds a test for the failing case

### DIFF
--- a/sqlalchemy_bigquery/base.py
+++ b/sqlalchemy_bigquery/base.py
@@ -838,7 +838,7 @@ class BigQueryDDLCompiler(DDLCompiler):
             field = time_partitioning.field
             if isinstance(
                 table.columns[time_partitioning.field].type,
-                sqlalchemy.sql.sqltypes.DATE,
+                (sqlalchemy.sql.sqltypes.DATE, sqlalchemy.sql.sqltypes.Date),
             ):
                 return f"PARTITION BY {field}"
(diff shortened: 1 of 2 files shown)


Linchin:
closing as we have merged #1113

PR #1122, opened 2024-09-30
chore: update templated files

Linchin:
Similar to https://github.com/googleapis/google-auth-library-python/pull/1597, this PR updates the template version, so owlbot updates can be triggered. I'm doing this because this release https://github.com/googleapis/python-bigquery-sqlalchemy/pull/1075 failed (successful commit, failed PyPI upload) is highly likely to be caused by an outdated PyPI key. With the key updated we can rerun the release job and it should be able to upload.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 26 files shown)


chalmerlowe (approved):
LGTM

chalmerlowe:
Owlbot failed. It didn't get past step one in the owlbot processing. That caused all the other steps to fail. I tried to run it locally ... there is a nox run command near the end of the owlbot.py file that fails cause it could not write (presumably inside the docker container the command is supposed to be running in).

In reviewing the code included in this PR, I am curious about the change to the `owlbot.lock.yaml` file.
I see that it is identical to the change in the PR you reference. Was your change manually generated by you OR automagically generated by a file update script as part of owlbot processing OR running the owlbot cli.

The reason I ask... I was under the impression that that file gets automagically updated via an update script, not manually and thus there may be other changes to the files.

Also... I notice there is no change to the `pypi` token identifier in this PR. Is that intentional. See this line in Anthonios' PR (I think the new token is `token 2`.): https://github.com/googleapis/google-auth-library-python/blob/8992bc9593173d3429d6ffd2651980e0ddf73dd7/.kokoro/release/common.cfg#L31

Linchin:
@chalmerlowe Thanks a lot for helping with the investigation! After talking with @ohmayr, it seems we had duplicated configs in `owlbot.py` and `setup.cfg`, and this led to owlbot being unable to update the script. After removing the copy in `owlbot.py`, as you can see owlbot added the updates it synced from the templates, which includes the key update.

PR #1155, opened 2025-01-19
fix: compilation of a labeled custom FunctionElement when used in grouping

### LINKED ISSUE #1150: Compilation of a labeled custom FunctionElement fails when used in grouping
Python 3.10.15
sqlalchemy_bigquery 1.12.0 (issue introduced in 1.11.0).

When creating a new FunctionalElement to compile some custom function that does not support the `default` dialect, and grouping by that element, we get an error as we are using the `default` dialect instead of compiler dialect.

#### Steps to reproduce
1. Create a new FunctionElement
2. Create a function that compiles it. Make sure it does not support the `default` dialect.
3. Create a query that groups by a field using this function and adding a label to it

#### Code example

```python
import sqlalchemy

class CustomLower(sqlalchemy.sql.functions.FunctionElement):
    name = "custom_lower"

@sqlalchemy.ext.compiler.compiles(CustomLower)
def compile_custom_intersect(element, compiler, **kwargs):
    if compiler.dialect.name != "bigquery":
        raise sqlalchemy.exc.CompileError(
            f"custom_lower is not supported for dialect '{compiler.dialect.name}'"
        )

    clauses = list(element.clauses)
    field = compiler.process(clauses[0], **kwargs)
    return f"LOWER({field})"

db_metadata = MetaData()

table1 = setup_table(
    engine,
    "table1",
    metadata,
    sqlalchemy.Column("foo", sqlalchemy.String),
    sqlalchemy.Column("bar", sqlalchemy.Integer),
)

lower_foo = CustomLower(table1.c.foo).label("some_label")
q = (
    sqlalchemy.select(lower_foo, sqlalchemy.func.max(table1.c.bar))
    .select_from(table1)
    .group_by(lower_foo)
)

print(q.compile(engine).string)
```

#### Error received
```
sqlalchemy.exc.CompileError: custom_lower is not supported for dialect 'default'
```

This is because when trying to understand if we have the `grouping_ops` in the element we use `str()` to stringify it instead of compiling it using the dialect. 

The solution should be in `visit_label`:
We should not use `str(column_label)`, but rather `column_label.compile(dialect=self.dialect).string`

#### Example fix
```python
    def visit_label(self, *args, within_group_by=False, **kwargs):
        # Use labels in GROUP BY clause.
        #
        # Flag set in the group_by_clause method. Works around missing
        # equivalent to supports_simple_order_by_label for group by.
        if within_group_by:
            column_label = args[0]
            sql_keywords = {"GROUPING SETS", "ROLLUP", "CUBE"}
            label_str = column_label.compile(dialect=self.dialect).string
            if not any(keyword in label_str for keyword in sql_keywords):
                kwargs["render_label_as_label"] = column_label

        return super(BigQueryCompiler, self).visit_label(*args, **kwargs)
```

snapiri:
If we have a labeled custom function that we are grouping by, and the function does not support the `default` dialect, we can not compile our query.

- [N/A] Appropriate docs were updated (if necessary)

Fixes #1150 🦕

### DIFF
--- a/tests/unit/test_compiler.py
+++ b/tests/unit/test_compiler.py
@@ -417,3 +417,41 @@ def test_complex_grouping_ops_vs_nested_grouping_ops(
     )
 
     assert found_sql == expected_sql
+
+
+def test_label_compiler(faux_conn, metadata):
+    class CustomLower(sqlalchemy.sql.functions.FunctionElement):
+        name = "custom_lower"
+
+    @sqlalchemy.ext.compiler.compiles(CustomLower)
+    def compile_custom_intersect(element, compiler, **kwargs):
+        if compiler.dialect.name != "bigquery":
+            raise sqlalchemy.exc.CompileError(
+                f"custom_lower is not supported for dialect {compiler.dialect.name}"
+            )
+
+        clauses = list(element.clauses)
+        field = compiler.process(clauses[0], **kwargs)
+        return f"LOWER({field})"
+
+    table1 = setup_table(
+        faux_conn,
+        "table1",
+        metadata,
+        sqlalchemy.Column("foo", sqlalchemy.String),
+        sqlalchemy.Column("bar", sqlalchemy.Integer),
+    )
+
+    lower_foo = CustomLower(table1.c.foo).label("some_label")
+    q = (
+        sqlalchemy.select(lower_foo, sqlalchemy.func.max(table1.c.bar))
+        .select_from(table1)
+        .group_by(lower_foo)
+    )
+    expected_sql = (
+        "SELECT LOWER(`table1`.`foo`) AS `some_label`, max(`table1`.`bar`) AS `max_1` \n"
+        "FROM `table1` GROUP BY `some_label`"
+    )
+
+    found_sql = q.compile(faux_conn).string
+    assert found_sql == expected_sql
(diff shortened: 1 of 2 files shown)


snapiri:
@tswast - I would appreciate any help with promoting this fix.

tests/unit/test_compiler.py
```diff
@@ -417,3 +417,41 @@ def test_complex_grouping_ops_vs_nested_grouping_ops(
...
+def test_label_compiler(faux_conn, metadata):
+    class CustomLower(sqlalchemy.sql.functions.FunctionElement):
+        name = "custom_lower"
+
+    @sqlalchemy.ext.compiler.compiles(CustomLower)
+    def compile_custom_intersect(element, compiler, **kwargs):
+        if compiler.dialect.name != "bigquery":
+            raise sqlalchemy.exc.CompileError(
+                f"custom_lower is not supported for dialect {compiler.dialect.name}"
+            )
```
tswast, line 431: This is failing our coverage.

```
tests/unit/test_compiler.py                  169      1      2      1    99%   429
```

Since we don't expect to actually hit this line, I think a `# pragma: NO COVER` would be OK.

```suggestion
            # We only test with the BigQuery dialect, so this should never happen.
            raise sqlalchemy.exc.CompileError(  # pragma: NO COVER
                f"custom_lower is not supported for dialect {compiler.dialect.name}"
            )
```

snapiri: Thanks!

tswast (approved):

tswast:
Thanks so much for the contribution @snapiri !

tswast:
This should be released as part of 1.12.1. Follow https://github.com/googleapis/python-bigquery-sqlalchemy/pull/1142 for further updates.

snapiri:
Thank you for pushing this forward @tswast !

PR #1186, opened 2025-04-03
test: adds pytest-xdist to nox sessions for experimentation

chalmerlowe:
DO NOT MERGE.
WIP experiments with pytest-xdist.


chalmerlowe:
Getting pytest-xdist to work with sqlalchemy is tricky. the _bigquery_create_db function needs to be populated so that it does its job.

chalmerlowe:
**Migration Notice:** This library is moving to the [google-cloud-python](https://github.com/googleapis/google-cloud-python) monorepo soon.

We closed this PR due to inactivity to ensure a clean migration. Please re-open this work in the new monorepo once the migration is complete!

PR #1200, opened 2025-05-09
fix: updates _get_transitive_schema_fields and tests

chalmerlowe:
Adds a line to the `_get_transitive_schema_fields` function to correct an issue with dot notation in column names.

Also adds a test suite of tests to exercise the new functionality.

Fixes internal issue #406556097 🦕

### DIFF
--- a/tests/unit/test__types.py
+++ b/tests/unit/test__types.py
@@ -3,6 +3,7 @@ from google.cloud.bigquery.schema import SchemaField
 
 from sqlalchemy_bigquery._types import _get_transitive_schema_fields, STRUCT_FIELD_TYPES
 
+
 def create_fut(name, field_type, mode="NULLABLE", sub_fields=None):
     """
     Helper function to create a SchemaField object for testing.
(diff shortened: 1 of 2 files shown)


tests/unit/test__types.py
```diff
@@ -0,0 +1,232 @@
...
+            create_schema_field_from_dict(sf_dict).to_api_repr()
+            for sf_dict in schema_dict.get("fields", [])
+        ],
+    }
+    return SchemaField.from_api_repr(api_repr)
+
+
+test_cases = [
+    (
+        "STRUCT field, not REPEATED, with sub-fields, should recurse",
```
tswast, line 39: [Optional] Generally, I use [pytest.param()](https://docs.pytest.org/en/latest/reference/reference.html#pytest-param) with an ID argument instead of a description. That makes it a little easier to select a particular test. See 

https://github.com/googleapis/python-bigquery-dataframes/blob/08d70b6ad3ab3ac7b9a57d93da00168a8de7df9a/tests/system/small/operations/test_plotting.py#L203-L204

for an example.

tswast (approved):
Love it. Thanks!

tests/unit/test__types.py
chalmerlowe: Good point, thanks for the reminder on this option.

PR #1215, opened 2025-07-21
feat: remove python 3.8 support

chalmerlowe:
* Removes Python 3.8 support because 3.8 is End of Life.
* Includes minor edits to support this change in configs and in the code

Also adds several minor tweaks that were useful for debugging some of the mods to this PR (and are useful in general for all future debugging efforts), such as:

* `pip freeze` to display what dependencies have been installed
~~* UPDATE: removed upon request: `@calculate_duration` decorator to confirm the total time to execute a `nox session`.~~


noxfile.py
```diff
@@ -56,11 +58,6 @@
     "tests",
 ]
 UNIT_TEST_EXTRAS_BY_PYTHON: Dict[str, List[str]] = {
-    "3.8": [
```
Linchin, line 59: Do we want to change this to 3.9 instead?

CONTRIBUTING.rst
```diff
@@ -148,7 +148,7 @@ Running System Tests
 
   .. note::
 
-      System tests are only configured to run under Python 3.8, 3.12, and 3.13.
+      System tests are only configured to run under Python 3.12, and 3.13.
```
Linchin, line 151: Need to add 3.9 here

chalmerlowe: Good catch, thanks! Fixed.

leahecole (approved):

PR #1225, opened 2025-07-29
chore: Update renovate.json

chalmerlowe:
Refine allowed versions for python 3.9 and geoalchemy2.
Renovate-bot keeps changing the version of geoalchemy2 to 0.18.0 and that only works with Python 3.10.


chalmerlowe:
OBE in deference to #1227

